PR #12, opened 2025-07-04
Allow optional and unknown strands.

Istar-Eldritch:
According to the "Table Detailing the Syntax for each Column" in the `Syntactical Requirements of GFF3 for INSDC – DRAFT 0.2`, the "Strand" column has the following specification: `[+-?.]{1}`. This indicates that it must contain one of the following characters: `+` (positive strand), `-` (minus strand), `.` (not stranded), or `?` (unknown strandedness). This PR makes so that `.` and `?` are correctly parsed when generating an embl file from gff3.

### DIFF
--- a/src/test/resources/gff3toff_rules/embl_strand.embl
+++ b/src/test/resources/gff3toff_rules/embl_strand.embl
@@ -1,0 +1,15 @@
+ID   1; SV XXX; XXX; XXX; XXX; XXX; 0 BP.
+XX
+AC   1;
+XX
+DE   .
+XX
+KW   .
+XX
+FH   Key             Location/Qualifiers
+FH
+FT   source          1..15215371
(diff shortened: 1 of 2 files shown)


src/test/resources/gff3toff_rules/unknown_strand.gff3
```diff
@@ -0,0 +1,3 @@
+##gff-version 3.1.26
+##sequence-region TEST01.1 1 6107
+TEST01.1	.	repeat_region	77	138	.	?	.	note=L2;
```
Rajkumar-D, line 3: Can you also check the following ENSEMBL region:
`1	ilAbrTril1.1	region	1	15215371	.	.	.	ID=region:1;Alias=OX276421.1`

PR #22, opened 2025-07-30
Validate ungrouped GFF3 features.

Istar-Eldritch:
Adds validation for non consecutive GFF3 features using the same sequence ID. While permitted in GFF3, this would lead to an EMBL file with multiple Entry for the same accession.


src/main/java/uk/ac/ebi/embl/converter/exception/UngroupedFeaturesException.java
```diff
@@ -0,0 +1,24 @@
...
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package uk.ac.ebi.embl.converter.exception;
+
+import lombok.Getter;
+import uk.ac.ebi.embl.converter.validation.ValidationRule;
+
+@Getter
```
Rajkumar-D, line 16: Please remove Getter in the Exception class

src/main/java/uk/ac/ebi/embl/converter/exception/UngroupedFeaturesException.java
```diff
@@ -0,0 +1,24 @@
...
+public class UngroupedFeaturesException extends ValidationException {
```
Rajkumar-D, line 17: Suggestion to rename class to
`public AnnotationAlreadyExistsException(int line, String offendingSeqId)`

src/main/java/uk/ac/ebi/embl/converter/gff3/reader/GFF3FileReader.java
```diff
@@ -44,6 +44,7 @@ public class GFF3FileReader implements AutoCloseable {
     GFF3Annotation currentAnnotation;
     String currentAccession;
     Map<String, GFF3SequenceRegion> accessionSequenceRegionMap = new HashMap<>();
+    HashSet<String> finishedAccessions = new HashSet<>();
```
Rajkumar-D, line 47: Suggestion to inject this HashSet to GFF3FileReader as follows
```
 // Set of converted accessions.
 HashSet<String> convertedAccessions = new HashSet<>();
  try (GFF3FileReader gff3Reader = new **GFF3FileReader(reader,convertedAccessions))** {
```
And then add it to the Set after returning from the readAnnotation()

while ((currentAnnotation = gff3Reader.readAnnotation()) != null) {
    ...
    if (isSameAnnotation(previousAnnotation, currentAnnotation)) {
        previousAnnotation.merge(currentAnnotation);
    } else {
        ....
        **convertedAccessions.add(previousAnnotation.getAccession());**
    }
}

Istar-Eldritch: but it breaks the encapsulation of the reader, now the callers needs to be aware of its internal logic.

src/main/java/uk/ac/ebi/embl/converter/exception/UngroupedFeaturesException.java
Istar-Eldritch: Renamed to DuplicateSeqIdException

Istar-Eldritch:
Addressed the naming comments. I do not think moving the HashSet outside of the reader is a good architectural choice.

The primary reason for this stance is **encapsulation and the clear responsibility of the `GFF3FileReader`**. The reader is not just a line-by-line parser; it's a GFF3 *file* reader. The GFF3 format has rules, and one of them, implicitly, is that sequence regions (identified by accession IDs) should be unique within the context of the annotations being processed from a single file stream. Enforcing this rule internally makes the `GFF3FileReader` a more robust and self-contained component.

Moving `processedAnnotations` out would shift a responsibility to the caller that is inherently tied to the correct interpretation and validation of the GFF3 file format. This would make the `GFF3FileReader` less "smart" about the format it's reading and could lead to:

Rajkumar-D (approved):
Please update the rule name to:
ValidationRule.GFF3_DUPLICATE_SEQUENCE_ID

PR #23, opened 2025-07-30
Validation rules documentation

Istar-Eldritch:
Adds documentation about validation rules how they work on this codebase and how to extend them.

[Rendered Document](https://github.com/enasequence/gff3tools/blob/34f278911114566d138a1615c9d9214c5b74388a/docs/0002_validation_rules.md)

### DIFF
--- a/docs/0002_validation_rules.md
+++ b/docs/0002_validation_rules.md
@@ -1,0 +1,202 @@
+- Feature Name: Validation Rules
+- Document Date: 2025-07-30
+- Last Updated: 2025-07-30
+
+# Summary
+
+This document describes the validation rules framework, a system designed to enforce data integrity and consistency across various data inputs within the GFF3Tools project. It provides a flexible and extensible mechanism for defining, applying, and reporting on validation rules, ensuring that data conforms to predefined standards and business logic.
+
+# Motivation & Rationale
+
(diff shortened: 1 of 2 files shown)


docs/0002_validation_rules.md
```diff
@@ -0,0 +1,202 @@
...
+
+The rationale behind the chosen design emphasizes extensibility, maintainability, and performance. A rules-based approach allows for easy addition of new validation checks without modifying core logic. The design prioritizes clear separation of concerns, making it easier to understand, debug, and enhance the framework.
+
+
+# System Overview / High-Level Design
+
+The validation rules framework is designed around a clear separation of concerns, with distinct components responsible for defining rules, managing their severity, and handling validation outcomes.
+
+**Main Components:**
+- **ValidationRule (uk.ac.ebi.embl.converter.validation.ValidationRule):** This enum serves as the central registry for all predefined validation rules. Each enum constant represents a specific validation check and includes a human-readable description.
```
Rajkumar-D, line 26: I assume that in the ValidationRule enum, we should also be able to configure the actual validation classes, for example:

```
GFF3_FEATURE_VALIDATION(
	ValidateCDSFeaturePhase.class,
	ValidateCDSFeatureTransalation.class,
)    
```

Each validation class will have a validate() method that contains the validation logic.

```
Class ValidateCDSFeaturePhase extends FeatureValidationRule{
   String validationMessage = "message"
    @Override
    public ValidationResult validate(Gff3Annotation gff3Annotation) {
        // Actual validation happens here.
    }
}
```

PR #37, opened 2025-09-29
ENA-6610-new

Rajkumar-D:


src/main/java/uk/ac/ebi/embl/gff3tools/gff3/reader/GFF3FileReader.java
```diff
@@ -22,13 +20,16 @@
...
+import uk.ac.ebi.embl.gff3tools.gff3.directives.GFF3Species;
 import uk.ac.ebi.embl.gff3tools.utils.Gff3Utils;
 import uk.ac.ebi.embl.gff3tools.validation.ValidationEngine;
 
 public class GFF3FileReader implements AutoCloseable {
 
     static Pattern VERSION_DIRECTIVE = Pattern.compile(
             "^##gff-version (?<version>(?<major>[0-9]+)(\\.(?<minor>[0-9]+)(:?\\.(?<patch>[0-9]+))?)?)\\s*$");
+    static Pattern SPECIES_DIRECTIVE =
+            Pattern.compile("^##species (?<taxonomyUrl>\\S*?[?&]name=(?<species>.*))\\s*$");
```
Istar-Eldritch, line 32: Is it always a taxonomyURL can it be just the species name? My concern is that this match will not cover all possible values, in some cases further processing may be required to extract the species or create the taxonomy from a species name.

PR #38, opened 2025-10-06
Code to handle ValidationWarning

Rajkumar-D:


src/main/java/uk/ac/ebi/embl/gff3tools/validation/ValidationEngine.java
lognath-ebi, line 69: We need handleValidationWarnings in the annotation level as well right ?

Istar-Eldritch (changes requested):
My understanding is that :

1. We want to allow to return multiple error types from the validations, some may be errors, some may be warnings.
2. We also want to disable the logic completely for performance reasons if its not to be used.

**This solution doesn't add any new or better way to address 1**. In the existing implementation in main, just returning a validation exception with a new id is enough for it to be enabled/disabled by the existing severity system using a configuration or cli input.
**This solution doesn't address 2**

PR #76, opened 2025-12-10
Ena 6648 fasta reader

IvaTutis:
A fasta reader that enables the user to read the sequences and stream/get as a string their slices. 

Currently measured to process a 2GB single fasta entry with a large sequence in about 2.8 sec - so that's reading, parsing, indexing, complaining if the file is malformed etc etc.
Getting small slices of the sequence is almost instantaneous, while streaming the entire sequence will of course take longer, but around 3 sec is to be expected. 

The user uses the reader through the interface in FastaFileService, the examples of usage are provided in FastaFileServiceIntegrationTest ("integration" because I am effectively testing the Fasta File Service together with the underlying SequentialFastaFileReader and the SequenceIndexBuilders and whatnot). 

Requiring feedback on: 
> suggestions for a better interface in FastaFileService
> structure
> suggestions for optimizations 
> potential spotted bugs 

The spec for the file I am reading is provided [here](https://docs.google.com/document/d/1Ez31iMUCRDvgDWm6vRYAa4foWRavDLcf55Z18PoCg6w/edit?tab=t.0#heading=h.pnsiydxse7el).

Currently I am aware of the following potential bugs:
> need check for trailing/leading Ns in case they span multiple lines 
> need to check if it's possible for there not to be a nextline character after the json in the header, and the sequence starts right after 
> probably should throw buffer sizes in some hardcoded constants file or appsettings


Rajkumar-D (commented):
Had initial comments on this PR.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/sequenceutils/SequenceIndexBuilder.java
```diff
@@ -0,0 +1,246 @@
...
+        return new Result(idx, s.nextHdr);
+    }
+
+    // =====================================================================
+    // =                          scanning core                            =
+    // =====================================================================
+
+    private static final class ScanState {
+        long pos; // absolute scan position
+        long firstBaseByte = -1; // first allowed base byte seen
```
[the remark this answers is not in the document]
IvaTutis: that's the lineFirstByte below

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/FastaEntryInternal.java
```diff
@@ -0,0 +1,27 @@
...
+package uk.ac.ebi.embl.gff3tools.fasta;
+
+import lombok.Getter;
+import lombok.Setter;
+import uk.ac.ebi.embl.gff3tools.fasta.headerutils.FastaHeader;
+import uk.ac.ebi.embl.gff3tools.fasta.sequenceutils.SequenceIndex;
+
+@Getter
+@Setter
+class FastaEntryInternal {
```
Rajkumar-D: Suggestion to rename this to FastaEntryMetadata
Can't we unify FastaEntry and FastaEntryMetadata?

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/FastaFileService.java
```diff
@@ -0,0 +1,154 @@
...
+
+    public List<FastaEntry> fastaEntries = new ArrayList<>();
+
+    private HashMap<String, SequenceIndex> sequenceIndexes = new HashMap<>();
+    private File file;
+    private SequentialFastaFileReader reader;
+
+    public FastaFileService() {
+        this.file = null;
+    }
```
Rajkumar-D: We can inject the file name into this constructor and make sure to call `openNewFile` to build the index only once.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/FastaFileService.java
```diff
@@ -0,0 +1,154 @@
...
+        close(); // if already open, close first
+        this.file = Objects.requireNonNull(fastaFile, "file");
+        this.fastaEntries.clear();
+        this.sequenceIndexes.clear();
+        reader = new SequentialFastaFileReader(fastaFile);
+        var readEntries = reader.readAll();
+        for (var entry : readEntries) {
+            FastaEntry fastaEntry = new FastaEntry();
+            fastaEntry.setSubmissionId(entry.getSubmissionId());
+            fastaEntry.setHeader(entry.getHeader());
```
Rajkumar-D: Couldn't we use the single object instead of using `FastaEntryInternal` and `FastaEntry`?

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/JsonHeaderParser.java
```diff
@@ -0,0 +1,109 @@
...
+        String normalized = raw.replace('\u201C', '"')
+                .replace('\u201D', '"')
+                .replace('\u2018', '\'')
+                .replace('\u2019', '\'')
+                .replace('\u00A0', ' ')
+                .trim();
+
+        JsonNode node;
+        try {
+            node = MAPPER.readTree(normalized);
```
Rajkumar-D: We can directly map JSON to java object using ObjectMapper.

```
ObjectMapper objectMapper = new ObjectMapper();
FastaHeader fastaHeader = objectMapper.readValue(jsonString, FastaHeader.class);
```

Rajkumar-D (commented):
Added comments

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/FastaHeader.java
```diff
@@ -0,0 +1,26 @@
...
+package uk.ac.ebi.embl.gff3tools.fasta.headerutils;
+
+import lombok.Getter;
+import lombok.Setter;
+import uk.ac.ebi.embl.gff3tools.fasta.Topology;
+
+@Getter
+@Setter
+public class FastaHeader {
+    String description; // mandatory
```
lognath-ebi: Its good keep it as private as we are using lombok setter it would make them publicly accessible.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/JsonHeaderParser.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new FastaFileException("Malformed FASTA header JSON. Failed to parse: " + normalized, e);
+        }
+
+        // Extract fields
+        Map<String, String> m = new HashMap<>();
+        node.fields().forEachRemaining(e -> {
+            String key = (e.getKey() == null ? "" : e.getKey())
+                    .trim()
+                    .toLowerCase(Locale.ROOT)
+                    .replaceAll("[\\s_-]+", "");
```
lognath-ebi: We won't need this logic if we use @JsonProperty at the field level of FastaHeader class.

lognath-ebi (dismissed):

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/FastaFileService.java
IvaTutis: i strongly prefer two, as the sequence indexes have to be sent to the fastafileservice somehow and it's easier to wrap it with the fasta entry header data in an object after the reading. 

just a subset of data is needed later on anyway.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/FastaHeader.java
IvaTutis: we need the setter though

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/JsonHeaderParser.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            return header;
+
+        } catch (JsonProcessingException e) {
+            throw new FastaFileException("Malformed FASTA header JSON: " + normalised, e);
+        }
+    }
+
+    private static String normaliseRawJsonString(String raw) {
```
Istar-Eldritch: I find this function not necessary. I just had a look around, no major json parser takes this stuff into consideration. I do not feel strongly about it though.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/sequenceutils/SequenceAlphabet.java
```diff
@@ -0,0 +1,62 @@
...
+    public String describeAllowed() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("[");
+
+        boolean first = true;
+        for (int i = 0; i < allowed.length; i++) {
+            if (allowed[i]) {
+                char c = (char) i;
+
+                // Render unprintables safely
```
Istar-Eldritch: All the nucleotide charaters are printable characters. This logic is unnecesary.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/sequenceutils/SequenceAlphabet.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /** Fast ASCII check for is it an allowed char. */
+    public boolean isAllowed(byte b) {
+        int i = b & 0xFF;
+        return i < 128 && allowed[i];
+    }
+
+    /** Fast ASCII check for 'N' or 'n' without decoding. */
+    public boolean isNBase(byte b) {
+        return ((b | 0x20) == 'n');
```
Istar-Eldritch: (Insert chef-kiss emoji here)
Not sure why but I didn't know / think of this trick before. And I've done a fair share of bit wizardry.

Istar-Eldritch (commented):
Looks good to me overall. I left a couple nits and I would like to look into it a bit deeper than I did. But I think as a first pass is very good.

I think this should be on its own repo that can be released as a library for anyone to use on their project. Also it would benefit a lot from some document explaining how to use it as a library.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/headerutils/JsonHeaderParser.java
```diff
@@ -0,0 +1,76 @@
...
+            .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES)
+            .build();
+    ;
+
+    public ParsedHeader parse(String headerLine) throws FastaFileException {
+        String rest = headerLine.substring(1);
+        int pipe = rest.indexOf('|');
+
+        String idPart = (pipe >= 0 ? rest.substring(0, pipe) : rest).trim();
+        String id = idPart.isEmpty() ? "" : idPart.split("\\s+")[0];
```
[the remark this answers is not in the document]
IvaTutis: Added a hard error if it's empty, as id should always be present. Good carch, rn it didn't do anything.

src/test/java/uk/ac/ebi/embl/gff3tools/fasta/FastaTestResources.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.*;
+import java.net.URL;
+import java.nio.file.*;
+import java.util.Objects;
+
+public final class FastaTestResources {
+    private FastaTestResources() {}
+
+    /** Returns a Path to a resource like ("fasta", "example.txt"). */
+    public static Path path(String dir, String fileName) {
```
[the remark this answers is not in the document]
IvaTutis: Realistically a good comment, but since we are planning to move it to a separate library I'll leave it separate.

src/main/java/uk/ac/ebi/embl/gff3tools/fasta/SequentialFastaFileReader.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /** Reads the next FASTA entry starting at or after 'from'. */
+    private Optional<FastaEntryMetadata> readNext(long from) throws FastaFileException {
+        try {
+            OptionalLong headerPosOpt = seekToNextHeader(from);
+            if (headerPosOpt.isEmpty()) return Optional.empty();
+
+            long headerPos = headerPosOpt.getAsLong();
+            String headerLine = readHeaderLine(headerPos);
```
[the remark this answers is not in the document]
IvaTutis: You are right on with the double read, since the original buffer during seek might catch the partial or whole header after the ">" in it with it. 

However, since the header line is limited to 4096 characters in that line according to the spec and I think that the separation of seek and reading of the header makes for an easier debug, I am moderately inclined to leave it as is. It's not much of a decline in terms of preformance.

PR #104, opened 2026-02-23
Ena 6839 sequence reading wrappers update

IvaTutis:
PR for the modification of readers. Basically the idea is like so: whatever type of file you have, you would open the file with the line:
`SequenceReader r = SequenceReaderFactory.ReadThisTypeOfFile(input params)`

This interface isn't exactly perfect as there are a couple of discrepancies between getting an unvalidated FASTA vs PlainSequence+JsonHeaderObject+accessionId, but essentially as long as the docs of the interface are followed we can have a singular entry point here.


src/main/java/uk/ac/ebi/embl/gff3tools/sequence/readers/PlainSequenceSubmissionReader.java
```diff
@@ -0,0 +1,100 @@
...
+import uk.ac.ebi.embl.gff3tools.sequence.IdType;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceStats;
+import uk.ac.ebi.embl.gff3tools.sequence.readers.fasta.header.utils.FastaHeader;
+
+public final class PlainSequenceSubmissionReader implements SubmissionSequenceReader {
+    private final SequenceReader sequenceReader;
+    private final String accessionId;
+    private final FastaHeader header; // nullable
+
+    public PlainSequenceSubmissionReader(File sequenceFile, String accessionId, FastaHeader optionalHeader)
```
Rajkumar-D: It is better to overload the constructor without optional argument

```
public PlainSequenceSubmissionReader(
        File sequenceFile,
        String accessionId) {
    this(sequenceFile, accessionId, null);
}
```

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SubmissionReaders.java
```diff
@@ -0,0 +1,94 @@
...
+ * <h2>Resource management</h2>
+ * <p>Readers returned by this factory hold file handles. Always close them when done.</p>
+ *
+ * <pre>{@code
+ * try (SubmissionSequenceReader reader = SubmissionReaders.openFasta(fastaFile)) {
+ *     // use reader...
+ * }
+ * }</pre>
+ */
+public final class SubmissionReaders {
```
Rajkumar-D: This could be SequenceReaderFactory?

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SubmissionReaders.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * <p>If accession IDs are supplied later, they must be provided in the same order as the records
+     * appear in the FASTA file.</p>
+     *
+     * @param fastaFile FASTA file to open (must not be {@code null})
+     * @return a {@link SubmissionSequenceReader} backed by the FASTA file
+     * @throws Exception if the file cannot be opened, parsed, or validated (e.g. invalid FASTA,
+     *                   invalid headers, duplicate IDs, I/O errors)
+     */
+    public static SubmissionSequenceReader openFasta(File fastaFile) throws Exception {
```
Rajkumar-D: Should this be readFasta()

Rajkumar-D (commented):
Added minor comments related to class/method names

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/readers/SequenceReader.java
```diff
@@ -0,0 +1,144 @@
...
+     *       otherwise implementations may return an empty list or throw (implementation choice; document it).</li>
+     * </ul>
+     *
+     * <p>For plain sequence:
+     * <ul>
+     *   <li>returns a single-element list containing the accession id for both id types (or throws for SUBMITTERID,
+     *       depending on your chosen semantics).</li>
+     * </ul>
+     */
+    List<String> getOrderedIds(IdType idType);
```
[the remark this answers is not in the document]
IvaTutis: It would generally return only submission or accession IDs in case of FASTA, as specified per input idType enum. 

In plain sequences, there ony exists one ID which is the accessionId (provided externally on input, as plain sequence files are generally allready validated beforehand and allready have their accessionId)

For plain sequences the function returns the same id regardless of input, which is incorrect but done for code convenience sake. It's documented in those lines. Theoretically for plain sequences it would be more accurate to return an empty list if a person demands the submission ids, and a list of one accession Id if the person demands accession id ordered list. 

The PlainSequenceReader is idType-agnostic in general, though the unified interface forces the difference

Rajkumar-D (approved):

PR #109, opened 2026-03-03
Implement accession substitution in the process command

Istar-Eldritch:
Add AccessionReplacementFix with feature-level and annotation-level methods to replace sequence IDs with ENA-assigned accessions. The fix is disabled in the validation engine (enabled=false) to avoid interfering with the reader's accession-based grouping, and is instead called as a post-processing step from FileProcessCommand. This will transition to engine-driven execution once context injection and priority execution features land.

FileProcessCommand.run() now parses accessions in two modes:
- Map mode (old:new pairs) for explicit mappings
- List mode (positional) with a first-pass GFF3 read for discovery

Output includes replaced accessions in sequence-region directives, feature column 1, and FASTA headers (with gzip support).


Rajkumar-D (commented):
Added a comment

Rajkumar-D (commented):
Could we enable AccessionReplacementFix during the post-processing step that updates the submitted accession with the INSDC accessions? If we could do that, we would not need to manipulate the writer.

Ideally, when running that, all the other validations/fixes could be ignored because at that point, the file would be valid.

Istar-Eldritch:
Closing, this will move to the internal repo

PR #116, opened 2026-03-19
Translate CDS features with multi-source sequence resolution and configurable output

Istar-Eldritch:
Changes:
  1. Add translate CLI command with FileSequenceProvider and TranslationFix
  2. Remove unused TranslationProvider and TranslationContext
  3. Docs updates
  4. Add composite sequence provider, output modes, and format detection
  5. E2E tests
  6. Support multiple --sequence with key:path


src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -0,0 +1,174 @@
...
+            String translation = result.getConceptualTranslation();
+            if (!translation.isEmpty()) {
+                for (GFF3Feature segment : segments) {
+                    segment.setAttributeList("translation", List.of(translation));
+                }
+                log.debug("Set translation attribute on CDS feature group at line {}", line);
+                recordTranslationState(representative, line, oldTranslation, translation);
+            }
+
+            propagateJoinAttributes(sorted);
```
Rajkumar-D, line 144: please move `propagateJoinAttributes(sorted);` inside to above condition block

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -0,0 +1,174 @@
...
+
+            if (result.hasErrors()) {
+                throw new ValidationException(
+                        "TRANSLATION", line, "Translation failed for CDS: " + result.getErrorMessages());
+            }
+
+            String translation = result.getConceptualTranslation();
+            if (!translation.isEmpty()) {
+                for (GFF3Feature segment : segments) {
+                    segment.setAttributeList("translation", List.of(translation));
```
Rajkumar-D: We could use  `GFF3Attributes.TRANSLATION` instead of   `"translation"`

please update all places

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -0,0 +1,174 @@
...
+                        sequenceLookup.getSequenceSlice(segment.getSeqId(), segment.getStart(), segment.getEnd());
+                concatenated.append(slice);
+            }
+
+            Translator translator = new Translator(representative);
+            translator.enableAllFixes();
+            TranslationResult result =
+                    translator.translate(concatenated.toString().getBytes());
+
+            if (result.hasErrors()) {
```
Rajkumar-D: Use `!result.isValid()` instead of `result.hasErrors()`

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -0,0 +1,174 @@
...
+            type = ANNOTATION,
+            priority = ValidationPriority.LOW)
+    public void fixAnnotation(GFF3Annotation annotation, int line) throws ValidationException {
+        if (!context.contains(SequenceLookup.class)) {
+            return;
+        }
+        SequenceLookup sequenceLookup = context.get(SequenceLookup.class);
+        if (sequenceLookup == null) {
+            return;
+        }
```
Rajkumar-D, line 87: This null check could be safely removed as there is the following check

`if (!context.contains(SequenceLookup.class)) {`

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/TranslationComparisonValidation.java
```diff
@@ -0,0 +1,96 @@
...
+public class TranslationComparisonValidation {
+
+    @InjectContext
+    private ValidationContext context;
+
+    @ValidationMethod(
+            rule = "TRANSLATION_COMPARISON",
+            description = "Compare existing translation with newly computed translation",
+            type = ANNOTATION,
+            severity = RuleSeverity.WARN,
```
Rajkumar-D: This should be `RuleSeverity.ERROR`

src/main/java/uk/ac/ebi/embl/gff3tools/cli/TranslationMode.java
```diff
@@ -0,0 +1,37 @@
...
+     */
+    public static class Converter implements CommandLine.ITypeConverter<TranslationMode> {
+        @Override
+        public TranslationMode convert(String value) {
+            String normalized = value.replace('-', '_');
+            try {
+                return TranslationMode.valueOf(normalized);
+            } catch (IllegalArgumentException e) {
+                throw new CommandLine.TypeConversionException(
+                        "expected one of " + Arrays.toString(TranslationMode.values()) + " but was '" + value + "'");
```
Rajkumar-D: Not fully shure if you need this, we could diside to accept only "-"

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
```diff
@@ -113,4 +111,64 @@ protected static Optional<String> getFileExtension(Path path) {
...
+     */
+    protected CompositeSequenceProvider buildCompositeProvider(
+            List<String> sequenceSpecs, SequenceFormat sequenceFormat) {
+        CompositeSequenceProvider compositeProvider = new CompositeSequenceProvider();
+        if (sequenceSpecs == null || sequenceSpecs.isEmpty()) {
+            return compositeProvider;
+        }
+        for (String spec : sequenceSpecs) {
+            ParsedSequenceSpec parsed = parseSequenceSpec(spec);
+            SequenceFormat format = resolveSequenceFormat(parsed.path(), sequenceFormat);
```
Rajkumar-D: The name `format` is too close to the parameter sequenceFormat. Rename to `resolvedFormat`

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
```diff
@@ -62,25 +63,22 @@ protected Map<String, RuleSeverity> getRuleOverrides() {
         return Optional.ofNullable(rules).map((r) -> r.rules()).orElse(new HashMap<>());
     }
 
-    protected ValidationEngine initValidationEngine(Map<String, RuleSeverity> ruleOverrides, Path processDir) {
+    protected ValidationEngine initValidationEngine(
+            Map<String, RuleSeverity> ruleOverrides, Path processDir, ContextProvider<?>... additionalProviders) {
```
Rajkumar-D: You can remove the code related to `processDir ` since it is not used.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/CompositeSequenceProvider.java
```diff
@@ -0,0 +1,75 @@
...
+public class CompositeSequenceProvider implements ContextProvider<SequenceLookup> {
+
+    private final List<SequenceSource> sources = new ArrayList<>();
+    private SequenceLookup cachedLookup;
+
+    /**
+     * Registers a sequence source. Sources are queried in registration order.
+     */
+    public void addSource(SequenceSource source) {
+        this.sources.add(source);
```
Rajkumar-D, line 37: Better to add a null check

`  Objects.requireNonNull(source, "source must not be null"); `

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
```diff
@@ -0,0 +1,157 @@
...
+        if (formatReader != null) {
+            try {
+                formatReader.close();
+            } catch (Exception e) {
+                log.warn("Failed to close sequence reader: {}", e.getMessage());
+            }
+        }
+    }
+
+    private void ensureInitialized() {
```
Rajkumar-D: Since we are running the conversion from the server, this method should be synchronized for handling concurrent threads.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/TranslationState.java
```diff
@@ -0,0 +1,47 @@
...
+     * Build a consistent lookup key for a feature.
+     *
+     * @param seqId the sequence ID
+     * @param featureId the feature ID, or null if absent
+     * @param line the GFF3 line number (used as fallback when featureId is null)
+     */
+    public static String buildKey(String seqId, String featureId, int line) {
+        String id = featureId != null ? featureId : "line_" + line;
+        return seqId + ":" + id;
+    }
```
Rajkumar-D, line 43: Suggestion to use the key from the following method:
`TranslationWriter.getTranslationKey(sequenceRegion.accession(), featureId.get()); `

src/main/java/uk/ac/ebi/embl/gff3tools/validation/ValidationContext.java
```diff
@@ -44,4 +48,32 @@ public <T> T get(Class<T> type) {
...
+        return providers.containsKey(type);
+    }
+
+    /** Close all registered providers, releasing any held resources. */
+    public void close() {
+        for (ContextProvider<?> provider : providers.values()) {
+            try {
+                provider.close();
+            } catch (Exception e) {
+                LOG.warn("Failed to close provider {}: {}", provider.type().getSimpleName(), e.getMessage());
```
Rajkumar-D, line 75: I think it is intentional to keep this WARN? not sure if there could be any issue

src/main/java/uk/ac/ebi/embl/gff3tools/validation/ValidationRegistry.java
```diff
@@ -91,13 +91,27 @@ private ValidationRegistry(ValidationConfig config, @Singular List<ContextProvid
...
+        Set<Class<?>> explicitTypes =
+                providers.stream().map(ContextProvider::type).collect(Collectors.toSet());
+        for (ContextProvider<?> provider : allProviders) {
+            if (!explicitTypes.contains(provider.type())) {
+                provider.initialize();
+            }
+        }
+        for (ContextProvider<?> provider : providers) {
+            provider.initialize();
+        }
```
Rajkumar-D, line 115: If the provider initialization fails, we should catch it.
```
 try {                                                                                                                                                                                                                                                                
              provider.initialize();      
          } catch (Exception e) {
              throw new IllegalStateException(                                                                                                                                                                                                                                 
                  "Failed to initialize provider " + provider.type().getSimpleName(), e);
          }   
```

Rajkumar-D (commented):
Assed few minor comments

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/TranslationState.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Shared state for old/new translation pairs, allowing {@code TranslationFix} to record
+ * translations and {@code TranslationComparisonValidation} to compare them.
+ */
+public class TranslationState {
+
+    public record TranslationEntry(String oldTranslation, String newTranslation) {}
+
+    private final Map<String, TranslationEntry> entries = new ConcurrentHashMap<>();
```
Rajkumar-D: TranslationFix writes and TranslationComparisonValidation reads sequentially, we could use HashMap instead of ConcurrentHashMap

src/main/java/uk/ac/ebi/embl/gff3tools/validation/ValidationContext.java
Istar-Eldritch: It is intentional, this is a best effort to close open resources, if it fails we still need to close all of the remaining ones, so we just return a warning. The other option would be to retain the errors and throw a final exception at the end. I do not htink this is required as this only happens once in the lifetime of the application.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
Istar-Eldritch: updated

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
Istar-Eldritch: updated

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
Istar-Eldritch: I kept this null check. `context.contains(SequenceLookup.class)` only confirms a **provider** is registered, but `CompositeSequenceProvider.get()` returns `null` when no sources have been added (e.g. when `--sequence` is not passed). Without this guard, we get an NPE at `sequenceLookup.getSequenceSlice(...)`.

The test `validationWithSequence_noSequence_stillSucceeds` confirms this, it runs validation without `--sequence` and expects success, but without the null check the CDS translation blows up with an NPE.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/TranslationComparisonValidation.java
Istar-Eldritch: done

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
Istar-Eldritch: done

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
Istar-Eldritch: done

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
Istar-Eldritch: done

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/TranslationState.java
Istar-Eldritch: done

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
Istar-Eldritch: I kept `propagateJoinAttributes` outside the `if` block. Moving it inside breaks pseudo gene handling, the `Translator` detects internal stop codons and sets the `pseudo` attribute, but may produce an empty translation string. The propagation of `pseudo` and 3' partiality to subsequent join segments must happen regardless of whether the translation is empty.

The existing test `propagatesPseudoToAllJoinSegments` confirms this: it mocks a sequence that triggers pseudo detection, and asserts that `pseudo` is propagated to seg2 — which fails if `propagateJoinAttributes` is gated on `!translation.isEmpty()`.

src/main/java/uk/ac/ebi/embl/gff3tools/cli/TranslationMode.java
Istar-Eldritch: Haha. Yes I've done this because I personally screw up with the -,_ when passing arguments to cli and I thought it was a good UX addition. Removed it.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/TranslationState.java
Istar-Eldritch: Done, but went a step further. The key logic (accession|urlEncode(featureId)) was duplicated between TranslationWriter and TranslationState. Instead of having one call the other, we
  extracted a shared TranslationKey.of(accession, featureId) utility in the gff3 package. Both TranslationWriter.getTranslationKey (now removed, callers use TranslationKey.of
  directly) and TranslationState.buildKey delegate to it.

  As part of this, TranslationState.buildKey now uses feature.accession() instead of feature.getSeqId(), aligning the key format across the codebase. The old line-number fallback for
  ID-less CDS features was removed, buildKey returns null when featureId is absent, and callers skip recording/lookup. The line-number fallback was unreliable anyway (annotation-level
   line number, not feature-level, so two ID-less CDS groups on the same annotation would collide).

Istar-Eldritch (commented):
addressed comments.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/TranslationState.java
Rajkumar-D: Ok, with this approach, please add null check for both in the method
```
Objects.requireNonNull(accession, "accession must not be null");                                                                                                                                                                          
Objects.requireNonNull(featureId, "featureId must not be null");
```

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/TranslationComparisonValidation.java
```diff
@@ -0,0 +1,107 @@
...
+            if (key == null) {
+                continue;
+            }
+            TranslationState.TranslationEntry translationEntry = state.get(key);
+            if (translationEntry == null || translationEntry.oldTranslation() == null) {
+                continue;
+            }
+            if (translationEntry.newTranslation() == null) {
+                log.debug("Skipping translation comparison for {}: old translation present but no new translation computed", key);
+                continue;
```
Rajkumar-D, line 93: We should remove the `continue` and throw an error when a new translation is not present

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/TranslationComparisonValidation.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            if (!translationEntry.oldTranslation().equals(translationEntry.newTranslation())) {
```
Rajkumar-D: We need to log when translation is successful, could be removed later
```
if (translationEntry.oldTranslation().equals(translationEntry.newTranslation())) {
                log.info("Successfully compared translation for: {}", key);
            }
```

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -0,0 +1,219 @@
...
+        if (hasException) {
+            return;
+        }
+
+        try {
+            // Concatenate sequence slices from all segments in genomic order
+            StringBuilder concatenated = new StringBuilder();
+            for (GFF3Feature segment : sorted) {
+                String slice =
+                        sequenceLookup.getSequenceSlice(segment.getSeqId(), segment.getStart(), segment.getEnd());
```
Rajkumar-D: Update `segment.getSeqId()` to `segment.accession() ` because the sequence will be saved using sequence + version

Rajkumar-D (commented):
Added minor comments
Please run spotless and commit

Rajkumar-D (approved):

PR #118, opened 2026-03-26
conflicts-resolved

IvaTutis:
Removed the sequence reader factory + specific set of readers, now they live in fasta reader as SequenceFormatReader and SequenceFormatReaderFactory

The header parsing has to be done in another place and not the fastareader lib reader


Rajkumar-D (approved):
Approved

PR #120, opened 2026-04-13
There was some error in translation while testing files shared by peter

Rajkumar-D:


src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -101,9 +101,14 @@ public void fixAnnotation(GFF3Annotation annotation, int line) throws Validation
 
     private void translateCdsGroup(List<GFF3Feature> segments, SequenceLookup sequenceLookup, int line)
             throws ValidationException {
-        // Sort segments by genomic start position
+
+        boolean isTransSpliced = segments.stream().anyMatch(s -> s.hasAttribute(GFF3Attributes.TRANS_SPLICING));
```
Istar-Eldritch, line 105: Good use of `anyMatch` here. Even a single trans-spliced segment means the whole join's order is intentional, so checking any segment in the group is the right approach. Maybe worth adding a comment?

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -113,6 +118,7 @@ private void translateCdsGroup(List<GFF3Feature> segments, SequenceLookup sequen
         boolean hasException = sorted.stream()
                 .anyMatch(s -> s.getAttribute(GFF3Attributes.EXCEPTION).isPresent());
         if (hasException) {
+            recordTranslationState(representative, line, oldTranslation, oldTranslation);
```
Istar-Eldritch, line 122: Good catch. Without this, features with exceptions (ribosomal slippage, etc.) would skip recording entirely and leave a gap in `TranslationState`. Setting `oldTranslation` as both old and new correctly reflects "no change was made." May be worth adding a comment explaining this?

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -158,24 +164,26 @@ private void translateCdsGroup(List<GFF3Feature> segments, SequenceLookup sequen
      * first (representative) feature. This method moves 3' partial to the last segment where
      * it semantically belongs, and propagates pseudo from the first segment to all others.
```
Istar-Eldritch: nit: this javadoc still says "moves 3' partial to the last segment" which describes the old behavior. Now the method does strand-aware partiality assignment using `TranslationResult` flags. Worth updating to match the new logic.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -158,24 +164,26 @@ private void translateCdsGroup(List<GFF3Feature> segments, SequenceLookup sequen
...
      */
-    private void propagateJoinAttributes(List<GFF3Feature> segments) {
+    private void propagateJoinAttributes(List<GFF3Feature> segments, TranslationResult result) {
         if (segments.size() <= 1) {
             return;
```
Istar-Eldritch: This early return means partiality flags from `TranslationResult` are never applied to single-segment CDS features. Since the partiality-setting code was removed from `Translator.translate()`, a single-segment CDS that triggers `isFixedFivePrimePartial()` or `isFixedThreePrimePartial()` (e.g. `codon_start != 1` making it 5' partial) will no longer get the partial attribute set anywhere.

Was this intentional? If single-segment CDS features should still get partiality, this early return needs to be removed or the partiality logic needs to be extracted above it.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFix.java
```diff
@@ -158,24 +164,26 @@ private void translateCdsGroup(List<GFF3Feature> segments, SequenceLookup sequen
...
-            if (!last.isThreePrimePartial()) {
-                last.setThreePrimePartial();
+        }
+
+        if (result.isFixedThreePrimePartial()) {
+            GFF3Feature threePrimeTarget = first.isComplement() ? first : last;
+            if (!threePrimeTarget.isThreePrimePartial()) {
+                threePrimeTarget.setThreePrimePartial();
             }
         }
```
Istar-Eldritch, line 189: The strand-aware target selection looks correct:

- `+` strand: 5' = first (lowest coord), 3' = last (highest coord)
- `-` strand: 5' = last (highest coord), 3' = first (lowest coord)

However, there is no test that exercises this complement-strand partiality propagation for a multi-segment join. Consider adding a test with a complement-strand multi-segment CDS where the translator flags 5' or 3' partial, then asserting the partial attribute lands on the correct segment.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/fix/TranslationFixTest.java
```diff
@@ -272,6 +273,42 @@ void recordsNullOldTranslationWhenNoPriorTranslation() throws Exception {
...
+        when(mockLookup.getSequenceSlice("seq1", 1L, 9L)).thenReturn("ATGAAAAAA");
+        when(mockLookup.getSequenceSlice("seq1", 10L, 18L)).thenReturn("CCCGGGTAA");
+
+        GFF3Feature segHigh = createFeature(OntologyTerm.CDS.name(), "cds1", "seq1", 10, 18, "+");
+        GFF3Feature segLow = createFeature(OntologyTerm.CDS.name(), "cds1", "seq1", 1, 9, "+");
+        // Add high-coord segment first (reverse genomic order)
+        GFF3Annotation annotation = createAnnotation(segHigh, segLow);
+        fix.fixAnnotation(annotation, 1);
+
+        // After sort: segLow(1-9) first → "ATGAAATAA"+"CCCGGGAAA" → translation "MKKPG"
```
Istar-Eldritch, line 288: nit: the inline comment says `"ATGAAATAA"+"CCCGGGAAA"` but the mocked sequences are `"ATGAAAAAA"` (1-9) and `"CCCGGGTAA"` (10-18), so the actual concatenation is `"ATGAAAAAA"+"CCCGGGTAA"`. The expected translation `"MKKPG"` is correct either way, but the comment is incorrect.

Istar-Eldritch (approved):
Some minor comments.

PR #122, opened 2026-04-13
ENA-6897: Add nucleotide sequence data to EMBL flatfile output

Istar-Eldritch:
Thread SequenceLookup from CompositeSequenceProvider through Gff3ToFFConverter into GFF3Mapper so that FASTA nucleotide data is populated on the Sequence object during GFF3-to-EMBL conversion. The EmblEntryWriter automatically writes the SQ section when sequence data is present.

### DIFF
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
@@ -10,6 +10,7 @@
  */
 package uk.ac.ebi.embl.gff3tools.gff3toff;
 
+import java.nio.ByteBuffer;
 import java.util.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -31,6 +32,7 @@ import uk.ac.ebi.embl.gff3tools.gff3.TranslationKey;
 import uk.ac.ebi.embl.gff3tools.gff3.directives.*;
 import uk.ac.ebi.embl.gff3tools.gff3.reader.GFF3FileReader;
 import uk.ac.ebi.embl.gff3tools.gff3.reader.OffsetRange;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
 import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.FastaHeaderProvider;
 import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.utils.FastaHeader;
 import uk.ac.ebi.embl.gff3tools.utils.ConversionEntry;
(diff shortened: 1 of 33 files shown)


src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
```diff
@@ -306,6 +313,30 @@ private Map<String, List<String>> getAttributesMap(GFF3Feature gff3Feature) {
...
+    private void applySequenceData(GFF3SequenceRegion sequenceRegion, Sequence sequence) {
+        if (sequenceLookup == null || sequenceRegion == null) {
+            return;
+        }
+        try {
+            String nucleotides = sequenceLookup.getSequenceSlice(
+                    sequenceRegion.accessionId(), sequenceRegion.start(), sequenceRegion.end());
+            // Convert to lowercase bytes in a single pass to avoid an intermediate
+            // String allocation from toLowerCase() (saves ~2N bytes of peak memory).
+            byte[] seqBytes = new byte[nucleotides.length()];
```
Rajkumar-D, line 342: Its good to check `nucleotides !=null && nucleotides.isNotEmpty()`

src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
```diff
@@ -306,6 +313,30 @@ private Map<String, List<String>> getAttributesMap(GFF3Feature gff3Feature) {
...
+            for (int i = 0; i < nucleotides.length(); i++) {
+                seqBytes[i] = (byte) Character.toLowerCase(nucleotides.charAt(i));
+            }
+            sequence.setSequence(ByteBuffer.wrap(seqBytes));
+            sequence.setLength((long) seqBytes.length);
+        } catch (Exception e) {
+            LOGGER.warn("Failed to retrieve sequence for '{}': {}", sequenceRegion.accessionId(), e.getMessage());
```
Rajkumar-D: Logging exception will silently be lost, throw valid exception so the pipeline will fail.

src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
```diff
@@ -306,6 +313,30 @@ private Map<String, List<String>> getAttributesMap(GFF3Feature gff3Feature) {
         return attributesMap;
     }
 
+    /**
+     * Populates the nucleotide sequence data on the Sequence object from the SequenceLookup.
+     * Gracefully skips if no lookup is available or no sequence region is defined.
+     */
+    private void applySequenceData(GFF3SequenceRegion sequenceRegion, Sequence sequence) {
+        if (sequenceLookup == null || sequenceRegion == null) {
+            return;
```
IvaTutis, line 329: add log.info maybe? i'd say warn but maybe they have only a gff3 and no sequence

src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
```diff
@@ -306,6 +313,30 @@ private Map<String, List<String>> getAttributesMap(GFF3Feature gff3Feature) {
...
+        }
+        try {
+            String nucleotides = sequenceLookup.getSequenceSlice(
```
IvaTutis, line 332: once main is merged into this branch, you'll have to add the SequenceRangeOption here. 

My knee-jerk reaction was to say "it will of course be the option which ignores edge N's", but that may be wrong as this is GFF3->FF converter, so either:
- we are going with a fixed-validated gff3 as a source that had it's N's stripped, in which case both options are equivalent
- orr the GFF3 supplied was not fixed for Ns (do we fix it before conversion?), in which case the annotation/sequence range will be off if we use the wrong option

Rajkumar-D (approved):

Rajkumar-D (changes requested):
Tests are failing, and need an error handling fix

src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
Istar-Eldritch: deferred until we parameterize this configuration as it should be applied consistently across the whole process, from validation and fixes to translation and conversion.

Rajkumar-D (approved):

PR #121, opened 2026-04-13
ENA-6897: Use JSON master file to generate EMBL entry during GFF3-to-EMBL conversion.

Istar-Eldritch:
Introduces a unified metadata pipeline for GFF3↔EMBL conversion. The headline architecture move is the consolidation of "annotation metadata" into MasterMetadata + MasterMetadataProvider (chain-of-responsibility), now wired through ValidationContext. GFF3Mapper.applyMasterMetadata() projects this onto Entry/Sequence/SourceFeature (DE, ID, AC, KW, CC, DT, OC, PR, RF, DR, source qualifiers). On the EMBL→GFF3 side, GFF3DirectivesFactory.createSpecies now reads from MasterMetadata instead of a master Entry, and FFToGff3Converter resolves it via the validation context. A parallel FastaHeaderProvider + FileFastaHeaderSource infrastructure plus a --fasta-header CLI option is added to expose FASTA-embedded headers and a CLI-supplied JSON fallback.

PR #123, opened 2026-04-14
ENA 6971: Add syntactic-ish validation for FastaHeader (mandatory values, etc)

IvaTutis:


IvaTutis:
will be adjusted after Rubens work with FastaHeaderProvider is done [ENA-6897: Map FASTA header JSON fields to EMBL entry during GFF3-to-EMBL conversion by Istar-Eldritch · Pull Request #121 · enasequence/gff3tools](https://github.com/enasequence/gff3tools/pull/121) to be injected into a high priority validation/fix

PR #127, opened 2026-04-27
TTENA-185:

Rajkumar-D:
TO fix the translation-related issue:
This branch is now focused on the joined-CDS translation

- TranslatorTest.java now uses the list-based constructor consistently.
- TranslationFixTest.java adds the regression for preserving 3' partiality across a joined CDS translation and makes that test self-contained by embedding the exact sequence slices inline.
- The old test that expected pseudo to be copied from the first segment to the rest of the join was removed, matching your intended behavior.

### DIFF
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
@@ -131,7 +131,7 @@ public class Translator {
         threePrimePartial = firstFeature.isThreePrimePartial() || lastFeature.isThreePrimePartial();
     }
 
-    private static GFF3Feature getFirstFeature(List<GFF3Feature> features) throws IllegalArgumentException{
+    private static GFF3Feature getFirstFeature(List<GFF3Feature> features) {
         if (features == null || features.isEmpty()) {
             throw new IllegalArgumentException("features must not be null or empty");
         }
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
@@ -86,9 +86,7 @@ public class Translator {
 
         feature = getFirstFeature(sortedFeatures);
 
-        fivePrimePartial = getCompoundFivePrimePartial(sortedFeatures);
-
-        threePrimePartial = getCompoundThreePrimePartial(sortedFeatures);
+        setCompoundPartiality(sortedFeatures);
 
         // TODO: Get translation table from taxon.
         int translationTable =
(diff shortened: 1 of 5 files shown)


src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
```diff
@@ -76,20 +76,26 @@ public class Translator {
...
+     * <p>Shared translation metadata is taken from the first segment, while join-level
+     * partiality is derived across all sortedFeatures so compound CDS translation preserves
+     * existing 5'/3' partial boundaries.
      */
-    public Translator(GFF3Feature feature) throws TranslationException {
+    public Translator(List<GFF3Feature> sortedFeatures) throws TranslationException {
+
+        feature = getFirstFeature(sortedFeatures);
+
+        fivePrimePartial = getCompoundFivePrimePartial(sortedFeatures);
```
lognath-ebi: Suggestion - Can we keep FivePrimePartial and ThreePrimePartial check in a single method ? I could see same operations being called twice get(0) and get(n-1) and also we fivePrimePartial, threePrimePartial in a  class level.

Rajkumar-D: Yes, moved it to a single method.

src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
```diff
@@ -76,20 +76,24 @@ public class Translator {
...
+     * Creates a translator for one or more CDS sortedFeatures.
      *
-     * @param feature the GFF3 feature associated with this translation
-     * @throws TranslationException if the translation table is invalid
+     * <p>Shared translation metadata is taken from the first segment, while join-level
+     * partiality is derived across all sortedFeatures so compound CDS translation preserves
+     * existing 5'/3' partial boundaries.
      */
-    public Translator(GFF3Feature feature) throws TranslationException {
+    public Translator(List<GFF3Feature> sortedFeatures) throws TranslationException {
```
Istar-Eldritch, line 85: We should add a global validation that checks the features are in order. This is out of scope for this PR, but worth adding a ticket for it.

src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
```diff
@@ -98,23 +102,39 @@ public Translator(GFF3Feature feature) throws TranslationException {
...
-
         codonStart = feature.getAttribute(GFF3Attributes.CODON_START)
                 .map(Integer::parseInt)
                 .orElse(codonStart);
 
         // Parse transl_except attributes and add position exceptions
         handleTranslExceptAttributes(feature);
 
-        // Parse codon attributes and add codon exceptions
         handleCodonExceptAttributes(feature);
```
Istar-Eldritch, line 113: I have the impression this comment is still useful?

src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
```diff
@@ -98,23 +102,39 @@ public Translator(GFF3Feature feature) throws TranslationException {
...
+        if (features.isEmpty()) {
+            fivePrimePartial = false;
+            threePrimePartial = false;
+            return;
+        }
+
+        GFF3Feature firstFeature = features.get(0);
+        GFF3Feature lastFeature = features.get(features.size() - 1);
+
+        fivePrimePartial = firstFeature.isFivePrimePartial() || lastFeature.isFivePrimePartial();
```
Istar-Eldritch, line 130: Worth a one-line comment explaining why ORing across both bookends is correct rather than checking strand explicitly. `GFF3Feature.isFivePrimePartial` and `isThreePrimePartial` are already strand-aware (they key off `start` vs `end` based on `isComplement`). On forward strand the 5' partial lives on the first segment; on complement strand it lives on the last. ORing is a strand-agnostic shortcut that works because only the strand-correct end carries the flag in valid input. Without a comment, a future reader will wonder if this is masking a bug.

src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
```diff
@@ -98,23 +102,39 @@ public Translator(GFF3Feature feature) throws TranslationException {
...
+        threePrimePartial = firstFeature.isThreePrimePartial() || lastFeature.isThreePrimePartial();
+    }
+
+    private static GFF3Feature getFirstFeature(List<GFF3Feature> features) {
+        if (features == null || features.isEmpty()) {
+            throw new IllegalArgumentException("features must not be null or empty");
```
Istar-Eldritch, line 136: This guard throws `IllegalArgumentException` (unchecked), but the constructor is declared `throws TranslationException`. Minor, but inconsistent. The only non-test caller (`TranslationFix.translateCdsGroup`) builds the list itself from a non-empty group so the guard is essentially defensive. Either drop it or align the exception type with the declared `throws`.

Istar-Eldritch (approved):
Approved with some very minor suggestions / comments.

src/main/java/uk/ac/ebi/embl/gff3tools/translation/Translator.java
Rajkumar-D: We can't change this to TranslationException as this is not a domain translation error. We can keep IllegalArgumentException

PR #128, opened 2026-04-29
TTENA-11:Adding trans_table in GFF3Attributes

Manishemblebi:


Istar-Eldritch (dismissed):

Istar-Eldritch (changes requested):
Not sure where this constant is used. If its not used as part of this codebase, why add it here?

PR #129, opened 2026-04-30
TTENA-190: Adding mapping for guide_RNA with matching attribute valie

Rajkumar-D:


Woolly-at-EBI (approved):
I fully agree with this change
Furthermore I think that we had an error in our mapping file anyway. (Sorry may have been my or less likely Josie's mistake.)
1)  The SO value is guide_RNA and not guideRNA
https://www.ebi.ac.uk/ols4/ontologies/so/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FSO_0000602?lang=en

2) In EMBL/GenBank there are only guide_RNA and no guideRNA
esearch -db nucleotide -query 'guide_RNA[All Fields] AND ncRNA[Feature key]' | \
    xtract -pattern ENTREZ_DIRECT -element Count
1590
esearch -db nucleotide -query 'guideRNA[All Fields] AND ncRNA[Feature key]' | \
    xtract -pattern ENTREZ_DIRECT -element Count
0

PR #131, opened 2026-05-08
ENA-6897: Introduce TaxonProvider SPI and segregate CLI options by workflow

Istar-Eldritch:
Introduces `TaxonProvider` as a lightweight SPI for taxon resolution, and cleans up how CLI options are distributed across commands.

## TaxonProvider SPI

Adds a `TaxonProvider` interface (`ContextProvider<TaxonProvider>`) with a single method `Optional<Taxon> resolve(long taxId)`. This is the integration point for downstream modules (e.g. `gff3-validations`) or the processing/ingestion pipeline to register a network-backed taxon resolver. Core ships no implementation — the tool remains offline by default.

Validators that need to resolve a taxon ID (e.g. to fetch a translation table) call `context.get(TaxonProvider.class).resolve(id)`, reading the ID directly from the annotation data rather than relying on a CLI-supplied value.

## CLI option segregation

`--master-entry` moves from `AbstractCommand` to `FileConversionCommand`, where it is only relevant (GFF3 to/from EMBL conversion). The validation workflow no longer exposes it.


src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
```diff
@@ -192,6 +206,63 @@ private Converter getConverter(
...
+    private MasterMetadataProvider buildMetadataProvider(Path masterEntryPath) throws ExitException {
+        MasterMetadataProvider provider = new MasterMetadataProvider();
+        if (masterEntryPath == null) {
+            return provider;
+        }
+        String ext = getFileExtension(masterEntryPath).orElse("").toLowerCase();
+        switch (ext) {
+            case "json" -> provider.addSource(parseMasterEntryJson(masterEntryPath));
+            case "embl", "ff" -> provider.addSource(new EmblEntryMetadataSource(parseMasterEntryEmbl(masterEntryPath)));
+            default ->
```
Rajkumar-D: Ruben, is this changes needed?

Istar-Eldritch: It is not functionally required for Jackson itself, but I added it to keep JSON and EMBL master-entry handling consistent and to return the same `NonExistingFile` CLI exception before parsing starts. If you prefer less duplication, I can remove the pre-check and rely on the existing `NoSuchFileException` catch.

src/main/java/uk/ac/ebi/embl/gff3tools/metadata/TaxonProvider.java
```diff
@@ -10,42 +10,26 @@
...
- * <p>The default provider has no sources and is offline-only. Network-backed sources
- * should be provided by environment-specific extensions and registered explicitly.
+ * <p>Core ships no implementation (offline by default). Online-backed implementations
+ * are registered by downstream modules (e.g. gff3-validations) or by the processing /
+ * ingestion pipeline at initialization time.
  */
-public class TaxonProvider {
-
-    private final List<TaxonSource> sources = new ArrayList<>();
+public interface TaxonProvider extends ContextProvider<TaxonProvider> {
```
Rajkumar-D, line 24: Should this be in the provider package? Do you need it to be an interface or a class?

Rajkumar-D (commented):
commented

src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
Istar-Eldritch: Follow-up: this code has since been refactored further and moved to `AbstractCommand.parseMasterEntryJson` / `parseMasterEntryEmbl`. The pre-existence check is still there in both methods for the reason I described above, to throw `NonExistingFile` (exit code 12) rather than `ReadException` (exit code 10) when the file is simply missing. Both JSON and EMBL paths are now symmetric.

src/main/java/uk/ac/ebi/embl/gff3tools/metadata/TaxonProvider.java
Istar-Eldritch: `gff3-validations` PR #8 implements `TaxonProvider` as a network-backed `ContextProvider` (`TaxonomyClientTaxonProvider`), so the interface is very much needed. 

To answer your original questions directly: it stays in the `metadata` package (that's where `gff3-validations` imports it from) and it needs to be an **interface**,  it's an SPI that downstream modules implement. Core ships no implementation; the tool is offline by default.

Rajkumar-D (approved):

PR #133, opened 2026-05-21
Ttena 231 232 238 tsv conversion fixes

IvaTutis:
An update to the TSV conversion: 
- added tests for all TSV templates when converting to GFF3 with FASTA with default header (public use)
- added option to create a GFF3 and JSON HEADER FASTA from TSV, tested for all templates and confimed validity
- added option to store SourceFeature data needed for the Master stage suring the TSV -> GFF3 conversion, confirmed re-readability (deseriasable)


src/main/java/uk/ac/ebi/embl/gff3tools/tsvconverter/TSVToGFF3Converter.java
```diff
@@ -63,8 +68,26 @@ public TSVToGFF3Converter(ValidationEngine validationEngine) {
      * @param fastaOutputPath optional output path for FASTA sequences; if null, sequences are discarded
      */
     public TSVToGFF3Converter(ValidationEngine validationEngine, Path fastaOutputPath) {
+        this(validationEngine, fastaOutputPath, FastaHeaderType.DEFAULT, null);
+    }
```
Rajkumar-D, line 72: Do we need this constructor? Please remove if not used

src/main/java/uk/ac/ebi/embl/gff3tools/utils/SourceFeatureDTO.java
```diff
@@ -0,0 +1,68 @@
...
+    public Taxon taxon;
+    public boolean focus;
+    public boolean transgenic;
+    public Map<String, String> qualifiers;
+
+    public SourceFeatureDTO() {}
+
+    public SourceFeatureDTO(SourceFeature original) {
+        this.scientificName = original.getScientificName();
+        this.commonName = original.getCommonName();
```
Rajkumar-D, line 49: Please do null on all fields check before setting

src/main/java/uk/ac/ebi/embl/gff3tools/utils/SourceFeatureDTO.java
```diff
@@ -0,0 +1,68 @@
...
+        this.qualifiers = new HashMap<>();
+
+        for (Qualifier qualifier : original.getQualifiers()) {
+            this.qualifiers.put(qualifier.getName(), qualifier.getValue());
+        }
+    }
+
+    public SourceFeature toSourceFeature() {
+        SourceFeature copy = new FeatureFactory().createSourceFeature();
+
```
Rajkumar-D, line 72: This method is not used now, I think the intention is to use it while master creation, Please check with Logan

Rajkumar-D (commented):
Commented

src/main/java/uk/ac/ebi/embl/gff3tools/utils/SourceFeatureDTO.java
IvaTutis: correct, it's to use there. Allready checked with him and made sure it matches what he does on master stage :)

src/main/java/uk/ac/ebi/embl/gff3tools/tsvconverter/TSVToGFF3Converter.java
IvaTutis: It is used by the CLI, I just made it fall back onto the bigger constructor which we will use in the pipeline

Istar-Eldritch (commented):
.

Istar-Eldritch (commented):
.

src/main/java/uk/ac/ebi/embl/gff3tools/utils/SourceFeatureDTO.java
```diff
@@ -0,0 +1,91 @@
...
+        this.transgenic = original.isTransgenic();
+
+        this.qualifiers = new HashMap<>();
+        if (original.getQualifiers() != null) {
+            for (Qualifier qualifier : original.getQualifiers()) {
+                if (qualifier == null) continue;
+
+                if (qualifier.getName() != null) {
+                    this.qualifiers.put(qualifier.getName(), qualifier.getValue());
+                }
```
Istar-Eldritch, line 60: Storing qualifiers as `Map<String, String>` makes this export/import lossy. Source features can carry multiple qualifiers with the same name, but `put()` keeps only the last one, so a round-trip through `SourceFeatureDTO` will silently drop the others.

Could we preserve qualifier multiplicity here, for example with `Map<String, List<String>>` or a list of name/value DTOs?

src/main/java/uk/ac/ebi/embl/gff3tools/tsvconverter/TSVToGFF3Converter.java
```diff
@@ -105,6 +144,10 @@ public void convert(BufferedReader reader, BufferedWriter writer)
             if (nucleotideFastaWriter != null) {
                 LOG.info("Wrote nucleotide sequences to FASTA file: {}", fastaOutputPath);
             }
+            if (sourceOutputPath != null) {
+                SourceFeatureUtils.dumpSourceFeatureDto(sourceFeatures, sourceOutputPath);
+                LOG.info("Wrote nucleotide sequences to FASTA file: {}", fastaOutputPath);
+            }
```
Istar-Eldritch, line 157: Small logging typo here: this branch writes the source-feature JSON, but the message still says `Wrote nucleotide sequences to FASTA file` and logs `fastaOutputPath`. It would be clearer to log `sourceOutputPath` and mention the source-feature output instead.

Istar-Eldritch (commented):
Thanks, this is a useful addition overall. I especially like the broader TSV template coverage and the JSON FASTA header tests.

I found one behavioral issue and one minor follow-up.

Istar-Eldritch (approved):

PR #134, opened 2026-06-05
TTNA-244: Add FASTA to GFF3 conversion emitting gap features

Istar-Eldritch:
Scan each FASTA sequence for contiguous runs of N bases and emit them as GFF3 gap features with estimated_length, gap_type, and linkage_evidence attributes. Supports gzip-compressed input via temp-file decompression.

- Add FastaToGff3Converter using fastareader GapRegion API
- Wire fasta->gff3 dispatch in FileConversionCommand
- Bump fastareader 1.1.0 -> 1.2.0 for gap-region support
- Add unit tests and fixture pair


src/main/java/uk/ac/ebi/embl/gff3tools/fftogff3/FastaToGff3Converter.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public void convert(BufferedReader reader, BufferedWriter writer)
+            throws ReadException, WriteException, ValidationException {
+
+        // The BufferedReader is not used; fastareader requires a File.
+        Path effectivePath = resolveGzippedPath(inputFilePath);
+
+        FileSequenceSource source = new FileSequenceSource(effectivePath, sequenceFormat, null);
```
IvaTutis: This is for internal use only,  but I would love to be able to extract information about the read file - sequence stats, indexes, so on. 

If possible, it would be ideal to feed this into the engine internally so that the sequence/annotation and some fastaheader format validations - the only validations that really make much sense in this case - can execute as they would in the FASTA+GFF3 input case. 

Unlike TSVs, FASTA submissions should make up a greater portion of our submissions and can be bigger files so I would like to avoid multiple reads.

src/test/java/uk/ac/ebi/embl/gff3tools/fftogff3/FastaToGff3ConverterTest.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+import uk.ac.ebi.embl.gff3tools.cli.SequenceFormat;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationEngine;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationEngineBuilder;
+
+class FastaToGff3ConverterTest {
```
IvaTutis, line 31: tests look good and pass for various id formats if I change it up to omit the . in FASTA submissionId. Could maybe add a test for that, but it works as is. 

Related to the previous comment, i would like to see a test that checks whether extraction of submissionIds(from eg. seqToOrdinal map)/sequenceInfo/fastaheader out of the context works once they are fed as usual.

IvaTutis (changes requested):
Would like to see feeding of the SequenceLookup and FastaHeader provider working here. Otherwise looks good, I think a converter makes more sense anyway rather than an empty input file with a flag.

src/test/java/uk/ac/ebi/embl/gff3tools/fftogff3/FastaToGff3ConverterTest.java
IvaTutis: Something like this, as a check

```java
ValidationContext context = engine.getContext();

ContextProvider<SequenceLookup> provider = context.getProvider(SequenceLookup.class);
if (!(provider instanceof CompositeSequenceProvider sequenceProvider)) {
    throw new IllegalStateException(
            "Registered provider for SequenceLookup is not a CompositeSequenceProvider");
}
CompositeSequenceProvider compositeSequenceProvider = (CompositeSequenceProvider) sequenceProvider;
//eg. check that there is only 1 source which is a FileSequenceSource w FASTA
// try getting the ids out of it

FastaHeaderProvider headerProvider = context.get(FastaHeaderProvider.class);
var fastaHeader = headerProvider.getHeader(someId);
//assert as expected
```

src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
```diff
@@ -64,6 +72,24 @@ public class FileConversionCommand extends AbstractCommand {
...
+            names = {"--min-gap-length", "-mgl"},
+            description = "Minimum run of N bases reported as a gap feature (FASTA to GFF3 conversion only). "
+                    + "Default: ${DEFAULT-VALUE}.")
+    public int minGapLength = FastaToGff3Converter.DEFAULT_MIN_GAP_LENGTH;
+
+    @CommandLine.Option(
+            names = {"--gap-type", "-gt"},
+            description = "Optional INSDC gap_type for generated gap features (FASTA to GFF3 conversion only). "
+                    + "When set, gaps map to assembly_gap; otherwise a plain gap is emitted.")
+    public String gapType;
```
Rajkumar-D, line 85: Do we need this? I think when converting to GFF3 we should always create `SO:0000730	gap`

src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
```diff
@@ -64,6 +72,24 @@ public class FileConversionCommand extends AbstractCommand {
...
+
+    @CommandLine.Option(
+            names = {"--linkage-evidence", "-le"},
+            description = "Optional INSDC linkage_evidence for generated gap features (FASTA to GFF3 conversion "
+                    + "only). Only valid with a gap_type that requires it (e.g. \"within scaffold\").")
+    public String linkageEvidence;
```
Rajkumar-D, line 91: Good that you cover linkage-evidence since it is available in some validation.

Rajkumar-D (commented):
commented

src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
Istar-Eldritch: the gap type is an INSCD thing, if we do not have a way to set it then the conversion to flatfile may not have all the expected metadata. I don't think it harms, but can be removed.

IvaTutis (approved):
Looked over and retested that information extraction works fine - thank you for the edits, works perfectly!

I think the optional flag in the CLI for gaps is fine. 

Bump up the gff3tools version and merge

PR #137, opened 2026-06-16
Support gzip-compressed sequence files in FileSequenceSource

Istar-Eldritch:
### What changed
- `FileSequenceSource` now transparently detects and decompresses `.gz` FASTA and plain sequence files to a temporary file before passing them to the `fastareader`.
- Added `GzipUtils` to share gzip detection and decompression logic between `FileSequenceSource` and `FileConversionCommand`.
- Removed the duplicate decompression logic in `FileConversionCommand` for the FASTA→GFF3 path; it now relies on `FileSequenceSource`.
- Temporary decompressed files are:
  - deleted deterministically in `FileSequenceSource.close()`,
  - registered with `deleteOnExit()` as a backstop if `close()` is not reached.
- Added tests for:
  - gzipped FASTA,
  - gzipped plain sequence,
  - uncompressed FASTA still working,
  - corrupt gzip failing cleanly without leaking a temp file.

### Why
The sequence provider (`--sequence`) previously failed on gzip-compressed FASTA files because the underlying `fastareader` library cannot read gzip directly. Users had to decompress files manually before running the tool.

### DIFF
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/utils/GzipUtils.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/utils/GzipUtils.java
@@ -51,7 +51,6 @@ public final class GzipUtils {
         Path tempFile = null;
         try {
             tempFile = Files.createTempFile(prefix, suffix);
-            tempFile.toFile().deleteOnExit();
             try (InputStream gzipIn = new GZIPInputStream(Files.newInputStream(source))) {
                 Files.copy(gzipIn, tempFile, StandardCopyOption.REPLACE_EXISTING);
             }
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/utils/GzipUtils.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/utils/GzipUtils.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2025 EMBL - European Bioinformatics Institute
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package uk.ac.ebi.embl.gff3tools.utils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.zip.GZIPInputStream;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
```diff
@@ -138,9 +158,32 @@ private synchronized void ensureInitialized() {
...
+            case plain -> SequenceFormatReaderFactory.readPlainSequence(filePath.toFile());
+        };
+    }
+
+    private Path resolvePath() throws NonExistingFile, ReadException {
+        Path filePath = decompressIfGzipped(path);
+        if (!filePath.equals(path)) {
+            decompressedPath = filePath;
+        }
+        return filePath;
```
Rajkumar-D, line 171: openReader() calls resolvePath() every time it is invoked. initialize() guards against double-initialization with initialized, but openReader() itself has no guard. If any future caller bypasses the initialized check, resolvePath() would decompress the file again, create a second temp file, overwrite decompressedPath with the new path, and orphan the first temp file permanently (since close() only tracks one path).

We could skip decompression if already done:     

```
  private Path resolvePath() throws NonExistingFile, ReadException {
      if (decompressedPath != null) {
          return decompressedPath;
      }
      Path filePath = decompressIfGzipped(path);
      if (!filePath.equals(path)) {                                                                                                                                                                                                                                                                                                                    
          decompressedPath = filePath;
      }
      return filePath;
  }

```

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
```diff
@@ -58,8 +62,17 @@ public class FileSequenceSource implements SequenceSource {
...
+    private volatile Path decompressedPath;
     private boolean initialized;
 
+    /**
+     * Returns the path to the temporary decompressed file, or null if the source
+     * was not gzipped. Exposed for tests only.
+     */
+    Path getDecompressedPathOrNull() {
+        return decompressedPath;
+    }
```
Rajkumar-D: We could use the @Getter instead?

src/main/java/uk/ac/ebi/embl/gff3tools/utils/GzipUtils.java
```diff
@@ -0,0 +1,70 @@
...
+     * Decompresses a gzip-compressed file to a temporary file.
+     *
+     * <p>The temporary file is registered with {@link File#deleteOnExit()} as a backstop so the
+     * JVM removes it on normal exit even if the caller's cleanup path is not reached.
+     */
+    public static Path decompressToTempFile(Path source, String prefix, String suffix) throws ReadException {
+        Path tempFile = null;
+        try {
+            tempFile = Files.createTempFile(prefix, suffix);
+            tempFile.toFile().deleteOnExit();
```
Rajkumar-D: deleteOnExit() adds the path to a static LinkedHashSet in the JVM that is never pruned: entries are only removed at JVM shutdown. When close() correctly deletes the temp file, the JVM still holds the path in that list for the rest of the process lifetime.

We could remove this line `tempFile.toFile().deleteOnExit();`

Rajkumar-D (commented):
Commented.

Rajkumar-D (approved):
Approved

PR #139, opened 2026-06-22
Ttena 180 sequence validations

IvaTutis:
**Prerequisites:**

- extended SequenceLookup interface
- SequenceLookup interface needs additional helper record classes, so I basically re-declared gff3tools equivalents of some fastareader classes for it, to make future devs not have to use fastareader if they want to make their own sequenceLookup/Source implementation - EDITED: REVERTED BACK TO FASTAREADER CLASSES AFTER DISCUSSION
- Added AssemblyContext and it's provider. AssemblyType is UNKNOWN as a default, and the min_gap_length is set to 10. This needs to be overriden programatically (while using gff3tools as a library), or with the CLI, which I did not add support for yet. 

**New validations (see [validation table](https://docs.google.com/spreadsheets/d/1WOoOg-k2vrWpiH4T2NHIcJHYSSH9inOy-0AwPBNPlTI/edit?usp=sharing))**

1. SequenceMappingValidation — Add validation that every feature accession resolves to a sequence when a sequence provider is registered, failing fast on unmapped accessions.
2. SequenceLengthValidation — Add validation enforcing minimum sequence-length thresholds (100 bp universal, 200 bp lncRNA, 1000 nt assembly) and ##sequence-region start/end consistency against the actual sequence length. This is meant to run after the fix that "cuts" off the sequences N-bases. **This required some of the old tests to be updated so that the sequences are larger.** 
3. FeatureLocationValidation — Add validation that feature and sequence-region coordinates stay within sequence bounds, rejecting end positions past the sequence length and start positions below 1.
4. GapFeatureBasesValidation — Add validation that gap/assembly_gap features span only contiguous N bases in the underlying sequence.
5. GapBasesPercentageValidation — Add validation that a non-chromosome sequence is not more than 50% N bases (excluding leading/trailing Ns), and never 100% Ns when it is a chromosome or lacks FASTA header info.
6. CdsTranslationPresenceValidation — Add validation that every translatable CDS receives a generated translation when translation was requested, exempting /exception, ID-less, and pseudo CDS.


src/main/java/uk/ac/ebi/embl/gff3tools/sequence/GapRegion.java
```diff
@@ -0,0 +1,32 @@
...
+    public long lengthBases() {
+        return endBase - startBase + 1;
+    }
+
+    public boolean overlaps(long fromBase, long toBase) {
+        if (fromBase < 1 || toBase < fromBase) {
+            throw new IllegalArgumentException("bad base range: " + fromBase + ".." + toBase);
+        }
+        return startBase <= toBase && endBase >= fromBase;
+    }
```
Rajkumar-D: Couldn't see the usage in gff3tools and validations, Please remove it if not used

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SequenceRangeOption.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Controls which portion of a sequence is considered when retrieving slices or gap regions.
+ */
+public enum SequenceRangeOption {
+    /** The full sequence, including any leading/trailing N bases at the edges. */
+    WHOLE_SEQUENCE,
+    /** The sequence with leading and trailing N bases excluded from consideration. */
+    WITHOUT_EDGE_N_BASES
+}
```
Rajkumar-D: I have seen a similar ENUM in FastaReader. Can't we use the same ENUM?

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SequenceStats.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package uk.ac.ebi.embl.gff3tools.sequence;
+
+import java.util.Map;
+
+public record SequenceStats(
```
Rajkumar-D: Same here can we use the class from Fastareader?

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/CdsTranslationPresenceValidation.java
```diff
@@ -0,0 +1,108 @@
...
+            if (isExempt(segments)) {
+                continue;
+            }
+            GFF3Feature representative = representativeOf(segments);
+
+            String key = TranslationState.buildKey(
+                    representative.accession(), representative.getId().orElse(null));
+            if (key == null) {
+                continue;
+            }
```
Rajkumar-D, line 72: The key shouldn't be null at this stage; if it is, something has gone wrong, and we should throw a hard error.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/CdsTranslationPresenceValidation.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private GFF3Feature representativeOf(List<GFF3Feature> segments) {
+        return segments.stream()
+                .min(Comparator.comparingLong(GFF3Feature::getStart))
+                .orElseThrow();
+    }
+
+    private boolean hasGeneratedTranslation(TranslationState.TranslationEntry entry) {
+        return entry != null && entry.newTranslation() != null;
```
Rajkumar-D: Add `&& !entry.newTranslation().isBlank();`

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+            throw new IllegalStateException("No sequence stats available for sequence " + seqId);
+        }
+
+        long lengthWithoutEdges = stats.totalBasesWithoutEdgeNBases();
+        long nCount = countInteriorNs(stats);
+        if (lengthWithoutEdges < 0 || nCount < 0) {
+            throw new IllegalStateException("Negative gap-base counts for sequence " + seqId + ": length="
+                    + lengthWithoutEdges + ", Ns=" + nCount);
+        }
+        if (lengthWithoutEdges == 0 || nCount == 0) {
```
Rajkumar-D, line 73: We should throw error case when lengthWithoutEdges = 0 since we have minimum length validation for the sequence

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+            return;
+        }
+
```
Rajkumar-D, line 76: Add comment
//For chromosomes, only check N count is less than the sequence count.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+        if (useLooseGapCheck(seqId)) {
+            if (nCount >= lengthWithoutEdges) {
+                throw new ValidationException(RULE_GAP_BASES_PERCENTAGE, line, MESSAGE_ALL_NS.formatted(seqId));
+            }
+        } else {
+            if (nCount > lengthWithoutEdges * MAX_GAP_FRACTION) {
```
Rajkumar-D, line 84: // N percentage check for non-chromosome sequences

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+        } catch (Exception e) {
+            throw new IllegalStateException("Unable to resolve sequence stats for sequence " + seqId, e);
+        }
+    }
+
+    private long countInteriorNs(SequenceStats stats) {
+        return stats.baseCount().getOrDefault('N', 0L) - stats.leadingNsCount() - stats.trailingNsCount();
+    }
+
+    private boolean useLooseGapCheck(String seqId) {
```
Rajkumar-D: rename -> isChromosomeSequence()

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidation.java
```diff
@@ -0,0 +1,69 @@
...
+            return;
+        }
+        SequenceLookup lookup = context.get(SequenceLookup.class);
+        long start = feature.getStart();
+        long end = feature.getEnd();
+        List<GapRegion> gapRegions;
+        try {
+            gapRegions = lookup.getGapRegions(feature.accession(), start, end, SequenceRangeOption.WHOLE_SEQUENCE);
+        } catch (Exception e) {
+            return;
```
Rajkumar-D: This will silently ignore a valid error case,  throw Validation so that the validation will fail.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidation.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        boolean covered = gapRegions.stream().anyMatch(g -> g.startBase() <= start && g.endBase() >= end);
```
Rajkumar-D: Should you check <= or < in this line?

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidation.java
```diff
@@ -0,0 +1,187 @@
...
+            type = ValidationType.ANNOTATION,
+            priority = ValidationPriority.NORMAL)
+    public void validateSequenceRegionAgainstSequence(GFF3Annotation annotation, int line) throws ValidationException {
+        GFF3SequenceRegion region = annotation.getSequenceRegion();
+        if (region == null) {
+            return;
+        }
+        long firstBaseIndex = 1L;
+        Long lastBaseIndex = resolveSequenceLength(annotation.getAccession());
+        if (lastBaseIndex == null) {
```
Rajkumar-D: Add comment
// When lastBaseIndex is null then sequence is not resolved from the ValidationContext

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidation.java
```diff
@@ -0,0 +1,187 @@
...
+        for (GFF3Feature feature : annotation.getFeatures()) {
+            Optional<String> soId = ontologyClient.findTermByNameOrSynonym(feature.getName());
+            if (soId.isPresent() && ontologyClient.isSelfOrDescendantOf(soId.get(), OntologyTerm.LNCRNA.ID)) {
+                return Optional.of(feature);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private Long resolveSequenceLength(String seqId) {
```
Rajkumar-D: Same method is also used in FeatureLocationValidation extract these to common method if possible

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/AnalysisContext.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.Getter;
+
+/**
+ * User-supplied analysis settings made available to fixes and validators via the
+ * validation context.
+ *
+ * <p>Both the {@link AnalysisType} type and the minimum gap size are external input. The
+ * minimum gap size must always be greater than zero.
+ */
+public final class AnalysisContext {
```
Rajkumar-D, line 23: Will this be set from the pipeline?

Rajkumar-D (commented):
Added comments:
Few major comments related to error handling.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/AnalysisContext.java
IvaTutis: yes, this should be set externally through the pipeline. min_gap is available from ENAPRO, as is analysis type

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidation.java
Istar-Eldritch: **[Must fix]** Swallowed exception makes the check a no-op on I/O failures.

When `lookup.getGapRegions(...)` throws (e.g. sequence file unreadable), this `catch` silently returns, making the feature appear valid. Every sibling validator (`GapBasesPercentageValidation`, `FeatureLocationValidation`) propagates the failure as an `IllegalStateException`. This should do the same — or better, let the exception surface so the caller can map it to the correct exit code.

```java
} catch (Exception e) {
    throw new IllegalStateException("Unable to fetch gap regions for " + feature.accession(), e);
}
```

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FeatureLocationValidation.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+        if (context.contains(SequenceLookup.class)) {
+            SequenceLookup lookup = context.get(SequenceLookup.class);
+            if (lookup != null) {
+                try {
+                    Long length = lookup.getSequenceLength(seqId, SequenceRangeOption.WHOLE_SEQUENCE);
+                    sequenceLengthCache.put(seqId, length);
+                    return length;
+                } catch (Exception ex) {
+                    throw new IllegalStateException("Unable to resolve sequence length for " + seqId, ex);
```
Istar-Eldritch: **[Should fix]** `IllegalStateException` bypasses the `ExitException`/`CLIExitCode` contract.

A lookup failure here exits with code 1 (GENERAL/unexpected) rather than 10 (READ_ERROR). If this is truly unreachable after `SequenceMappingValidation` runs at HIGH priority, document that assumption with a comment. If it can happen, wrap in `ReadException` or equivalent.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FeatureLocationValidation.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String FEATURE_END_EXCEEDS_SEQUENCE_LENGTH =
+            "The end position of the location \"%s\" is greater than the length of the sequence (\"%d\").";
+
+    private static final String RULE_FEATURE_START_BELOW_ONE = "FEATURE_START_BELOW_ONE";
+    private static final String FEATURE_START_BELOW_ONE = "The start position of the location \"%s\" is less than 1.";
+
+    @InjectContext
+    private ValidationContext context;
+
+    private final Map<String, Long> sequenceLengthCache = new HashMap<>();
```
Istar-Eldritch, line 42: **[Should fix]** Instance-level cache — lifecycle depends on how validators are instantiated.

If `FeatureLocationValidation` instances are reused across multiple input files in one process, (like in the case of the gff3 web service) this cache carries over stale sequence lengths from previous files. If instances are always created fresh per file, this is safe (like in the case of the pipeline or cli). Because of the web impl we need to make it stateless by moving the cache into `ValidationContext` or some other mechanism.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidation.java
```diff
@@ -0,0 +1,187 @@
...
+import uk.ac.ebi.embl.gff3tools.exception.ValidationException;
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Annotation;
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.gff3.directives.GFF3SequenceRegion;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceRangeOption;
+import uk.ac.ebi.embl.gff3tools.utils.OntologyClient;
+import uk.ac.ebi.embl.gff3tools.utils.OntologyTerm;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+import uk.ac.ebi.embl.gff3tools.validation.meta.*;
```
Istar-Eldritch, line 24: Nit: wildcard import (`validation.meta.*`) while every other file in the package uses explicit imports. `spotlessApply` may flag this — worth aligning regardless.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidation.java
```diff
@@ -0,0 +1,187 @@
...
+    private static final String MESSAGE_LNCRNA_TOO_SHORT =
+            "lncRNA sequences usually have a length greater than 200bp. Please check that you are certain about this annotation.";
+    private static final String MESSAGE_ASSEMBLY_SEQUENCE_TOO_SHORT =
+            "Sequence assembly submissions require sequences to be at least 1000 nt long.";
+
+    @InjectContext
+    private ValidationContext context;
+
+    private final Map<String, Long> sequenceLengthCache = new HashMap<>();
+    private Boolean sequenceAssembly;
```
Istar-Eldritch, line 62: **[Should fix]** Same instance-level cache concern as `FeatureLocationValidation`.

`sequenceLengthCache` and especially `sequenceAssembly` (line 64) are instance fields. `sequenceAssembly` is computed once via `isSequenceAssembly()` and never reset — if the validator is reused across files with different analysis contexts, the cached value will suppress assembly-length checks on subsequent runs.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidation.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        if (context.contains(SequenceLookup.class)) {
+            SequenceLookup lookup = context.get(SequenceLookup.class);
+            if (lookup != null) {
+                try {
+                    Long length = lookup.getSequenceLength(seqId, SequenceRangeOption.WITHOUT_EDGE_N_BASES);
+                    sequenceLengthCache.put(seqId, length);
+                    return length;
+                } catch (Exception ex) {
+                    throw new IllegalStateException("Unable to resolve sequence length for " + seqId, ex);
```
Istar-Eldritch: Same `IllegalStateException` vs `ExitException` concern as in `FeatureLocationValidation` — see comment there.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final String MESSAGE_ALL_NS =
+            "Sequence \"%s\" consists entirely of N bases (excluding leading/trailing Ns).";
+
+    @InjectContext
+    private ValidationContext context;
+
+    @ValidationMethod(
+            rule = RULE_GAP_BASES_PERCENTAGE,
+            description = "A sequence (excluding edge Ns) must not be more than 50% Ns, unless it is a chromosome or no"
+                    + " FASTA header information is available, in which case it must only not be 100% Ns",
```
Istar-Eldritch, line 49: **[Nit]** Description does not match the implementation of `useLooseGapCheck`.

The description says "no FASTA header information is available → loose check". But `useLooseGapCheck` returns `false` (strict) when a `FastaHeaderProvider` is registered but returns no header for the given `seqId`. Only the complete absence of a `FastaHeaderProvider` triggers the loose check. If this distinction is intentional, the description should say so explicitly.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        SequenceLookup lookup = context.get(SequenceLookup.class);
+        if (lookup == null) {
+            return;
+        }
+
+        String seqId = annotation.getAccession();
+        SequenceStats stats = resolveStats(lookup, seqId);
+        if (stats == null) {
+            throw new IllegalStateException("No sequence stats available for sequence " + seqId);
```
Istar-Eldritch, line 64: Same `IllegalStateException` vs `ExitException` concern — a stats lookup failure will exit with code 1 instead of 10.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceMappingValidation.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final String RULE_SEQUENCE_MAPPING = "SEQUENCE_MAPPING";
+    private static final String NO_MAPPING_MESSAGE =
+            "No sequence could be resolved for accession \"%s\". A sequence provider is registered but does not "
+                    + "supply a sequence for this feature.";
+
+    @InjectContext
+    private ValidationContext context;
+
+    private final Set<String> validatedAccessions = new HashSet<>();
```
Istar-Eldritch, line 42: **[Should clarify]** Same instance-level cache concern — `validatedAccessions` persists across files if the instance is reused.

Also, on a lookup failure `resolveOrExplode` removes the accession from the set before throwing (line 70), allowing a retry to re-run the lookup. This is probably correct, but a brief comment would make the intent explicit.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
```diff
@@ -102,10 +108,71 @@ public boolean hasSequence(String seqId) {
...
+    }
+
+    @Override
+    public long getSequenceLength(String seqId, SequenceRangeOption option) throws Exception {
+        ensureInitialized();
+        long ordinal = resolveOrdinal(seqId);
+        uk.ac.ebi.embl.fastareader.SequenceStats fr = formatReader.getStats(ordinal);
+        return switch (option) {
+            case WHOLE_SEQUENCE -> fr.totalBases();
+            case WITHOUT_EDGE_N_BASES -> fr.totalBasesWithoutNBases();
```
Istar-Eldritch, line 125: **[Verify]** Confirm that `totalBasesWithoutNBases()` excludes only leading/trailing Ns, not all interior Ns.

`SequenceLengthValidation.resolveSequenceLength` uses `WITHOUT_EDGE_N_BASES` when applying the 100/200/1000 nt thresholds. If the library method strips all Ns (not just edge ones), sequences could incorrectly fail or pass the minimum-length checks.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceLengthValidationTest.java
```diff
@@ -0,0 +1,296 @@
...
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Annotation;
+import uk.ac.ebi.embl.gff3tools.gff3.directives.GFF3SequenceRegion;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceRangeOption;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+import uk.ac.ebi.embl.gff3tools.validation.provider.AnalysisContext;
+import uk.ac.ebi.embl.gff3tools.validation.provider.AnalysisType;
+
+public class SequenceLengthValidationTest {
```
Istar-Eldritch: By convention test classes should be package-private — remove the `public` modifier.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidationTest.java
```diff
@@ -0,0 +1,109 @@
...
+import uk.ac.ebi.embl.gff3tools.TestUtils;
+import uk.ac.ebi.embl.gff3tools.exception.ValidationException;
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.sequence.GapRegion;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceRangeOption;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+
+public class GapFeatureBasesValidationTest {
```
Istar-Eldritch: Nit: same — `public class` should be package-private.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FeatureLocationValidationTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.jupiter.api.Test;
+import uk.ac.ebi.embl.gff3tools.TestUtils;
+import uk.ac.ebi.embl.gff3tools.exception.ValidationException;
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceRangeOption;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+
+public class FeatureLocationValidationTest {
```
Istar-Eldritch: Nit: same — `public class` should be package-private.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidationTest.java
```diff
@@ -0,0 +1,229 @@
...
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Annotation;
+import uk.ac.ebi.embl.gff3tools.gff3.directives.GFF3SequenceRegion;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceStats;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.FastaHeaderProvider;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.utils.FastaHeader;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+
+public class GapBasesPercentageValidationTest {
```
Istar-Eldritch: Nit: same — `public class` should be package-private.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/SequenceMappingValidationTest.java
```diff
@@ -0,0 +1,122 @@
...
+import org.junit.jupiter.api.Test;
+import uk.ac.ebi.embl.gff3tools.TestUtils;
+import uk.ac.ebi.embl.gff3tools.exception.ValidationException;
+import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceRangeOption;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
+
+public class SequenceMappingValidationTest {
```
Istar-Eldritch: Nit: same — `public class` should be package-private.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidation.java
IvaTutis: it should be <=, gap regions have Ns at their edges. Inclusive interval on both ends.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
IvaTutis: It should and will fail on that validation rule instead then. No use in validating multiple things on one validation.

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SequenceRangeOption.java
IvaTutis: adjusted to use fastareader class

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SequenceStats.java
IvaTutis: adjusted to use fastareader class

Istar-Eldritch (commented):
Two issues must be fixed before merge; the rest are lower priority.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/CdsTranslationPresenceValidation.java
```diff
@@ -0,0 +1,115 @@
...
+            if (isExempt(segments)) {
+                continue;
+            }
+            GFF3Feature representative = representativeOf(segments);
+
+            String key = TranslationState.buildKey(
+                    representative.accession(), representative.getId().orElse(null));
+            if (key == null) {
+                throw new IllegalStateException(
+                        "Cannot build a translation key for CDS without an ID on accession '%s' at %d-%d"
```
Rajkumar-D, line 67: The key shouldn't be null at this stage; if it is, something has gone wrong, and we should throw a hard error.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/provider/FileSequenceSource.java
IvaTutis: Confirming:
- the totalBasesWithoutNBases is a fastareader typo, should be without edgeNbases. It is easy to confirm with how it's calculated in fastareader as totalBases-leadingNs-trailingNs
- the validation should use without edge n bases, confirmed at the bioinformatician meeting last week

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapFeatureBasesValidation.java
IvaTutis: addressed. will map to exit code one as if it happens, it;s our fault. see comment below.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FeatureLocationValidation.java
IvaTutis: addressed. will map to exit code 1. see comment below.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/CdsTranslationPresenceValidation.java
IvaTutis: addressed. will map to exit code 1 see comment below.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FeatureLocationValidation.java
IvaTutis: added exit method to all caches i found, including in implemented validations to clear caches upon end of file processing.

IvaTutis: addressed. added exit clearing of caches on validation which use caches

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/GapBasesPercentageValidation.java
IvaTutis: edited since. The rule is for anything that is 100% not a chromosome, we consider the stricter condition here. Otherwise if data is not available or not submitted at all, we use the lower baseline check as it ""might be a chromosome""

Istar-Eldritch (approved):

Rajkumar-D (approved):

PR #141, opened 2026-06-25
Ttena 258 fastaheader fixes and moved validations

IvaTutis:
Moves GFF3 validations into this package and adds new FASTA-header fixes and validations. See the [[tracking sheet](https://docs.google.com/spreadsheets/d/1WOoOg-k2vrWpiH4T2NHIcJHYSSH9inOy-0AwPBNPlTI/edit?gid=1978204027#gid=1978204027)](https://docs.google.com/spreadsheets/d/1WOoOg-k2vrWpiH4T2NHIcJHYSSH9inOy-0AwPBNPlTI/edit?gid=1978204027#gid=1978204027) for the full list.

> **Note for reviewers:** the moved validations, their tests, and the controlled-vocabulary utils were *already reviewed elsewhere* and **can be skipped**. Focus review on the new fixes/validations and the `excludeProvider` wiring below.

## Moved code

- Moved validations + their tests + controlled-vocabulary utils over from the gff3 validations package.

## New fixes

- **`FastaHeaderNormalisationFix`** — *high priority*, runs as an annotation fix. Makes the FASTA header ASCII-compliant (as in `sequencetools`), then normalises the controlled-vocabulary fields (`topology`, `mol_type`, `chrom_location`, `chrom_type`) to the expected format. *Confirmed with Colman that we **should** do this* to give submitters some flexibility.
- **`ChromosomeNameFix`** — additional cleaning approved for **chromosome names only**:
  - trims **all** whitespace (not just edges)
  - removes certain words and ASCII alphabet signs
  - collapses duplicate dashes
  - etc.

## New validation

- **`FastaHeaderMappingValidation`** *(new annotation validation)* — raises an error when a `FastaHeaderProvider` **is** defined in the validation context but the header is unavailable. This lets all other fixes/validations simply `log.warn` when the provider is present but no header exists for a given annotation.
- *Added a test to every validation that uses the `FastaHeaderProvider`*, confirming it **does not throw** when the provider is `undefined`.

## FastaHeaderProvider auto-exclusion when not defined 

- Added an **`excludeProvider`** builder method to the validation engine to remove a single provider and prevent it from auto-registering (used in tests):

  ```java
  engineBuilder.excludeProvider(FastaHeaderProvider.class);
  ```

- `AbstractCommand.initValidationEngine` now auto-calls the above when no provider is explicitly supplied. As a result, FASTA-header-related fixes/validations only run *fully* when a CLI command supplies a provider with a real header source — otherwise they auto-return, since the relevant context is absent from the `ValidationContext`.
- Applied the same change to **`FileConversionCommand`** — *needs review*.
- Updated tests to use `excludeProvider` where applicable, **or** supplied a fully-defined `FastaHeaderProvider` (fixing the headers in those cases).

## Additional
Switched all FastaHeader-specific validations and fixes to critical. This is neccessary as the trimEdgeNs fix will need to query topology and and all other validations should work with those changed coordinates - so following the chain of dependedncies, if fastaheader is defined poorly in the FASTA file, that should blow up before the trimEdgeN fix and following validations.

### DIFF
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
@@ -73,8 +73,19 @@ public abstract class AbstractCommand implements Runnable {
         ValidationEngineBuilder builder =
                 new ValidationEngineBuilder().overrideMethodRules(ruleOverrides).failFast(failFast);
 
+        boolean suppliedFastaHeaderProvider = false;
         for (ContextProvider<?> provider : additionalProviders) {
             builder.withProvider(provider);
+            if (provider.type() == FastaHeaderProvider.class) {
+                suppliedFastaHeaderProvider = true;
+            }
+        }
+
+        // When no caller supplies a FASTA header source, keep the autodetected (empty)
+        // FastaHeaderProvider out of the context. Otherwise header-aware rules such as
+        // FASTA_HEADER_MAPPING would fire against accessions for which no header was ever provided.
+        if (!suppliedFastaHeaderProvider) {
+            builder.excludeProvider(FastaHeaderProvider.class);
         }
 
         return builder.build();
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
@@ -73,19 +73,12 @@ public abstract class AbstractCommand implements Runnable {
         ValidationEngineBuilder builder =
                 new ValidationEngineBuilder().overrideMethodRules(ruleOverrides).failFast(failFast);
 
-        boolean suppliedFastaHeaderProvider = false;
+        // Providers gate their own registration via ContextProvider#isActive(). An empty
+        // FastaHeaderProvider (no header source supplied) reports inactive and is kept off the
+        // context, so header-aware rules such as FASTA_HEADER_MAPPING stay inert unless a caller
+        // supplies a real header source.
         for (ContextProvider<?> provider : additionalProviders) {
             builder.withProvider(provider);
-            if (provider.type() == FastaHeaderProvider.class) {
-                suppliedFastaHeaderProvider = true;
-            }
-        }
-
-        // When no caller supplies a FASTA header source, keep the autodetected (empty)
-        // FastaHeaderProvider out of the context. Otherwise header-aware rules such as
-        // FASTA_HEADER_MAPPING would fire against accessions for which no header was ever provided.
-        if (!suppliedFastaHeaderProvider) {
-            builder.excludeProvider(FastaHeaderProvider.class);
         }
 
         return builder.build();
--- a/src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
+++ b/src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
@@ -1061,15 +1061,12 @@ public class GFF3Mapper {
     }
 
     /**
-     * Maps chromosome_location to the EMBL /organelle qualifier on the source feature. Per the ENA
-     * assembly submission docs, a chromosome is assumed nuclear (no /organelle qualifier) when this
-     * field is absent; "nuclear" is handled as a case-insensitive synonym for absent here as a
-     * defensive fallback for inputs (e.g. the legacy MasterMetadata/master-JSON path) that were not
-     * normalised by {@code FastaHeaderNormalisationFix}, which strips it to null for the FastaHeader
-     * path. All other values are matched case-insensitively against the INSDC {@code /organelle}
-     * controlled vocabulary ({@link ControlledVocabularyUtils.ChromosomeLocation}); this tool now
-     * owns validation of that vocabulary, so unrecognised values are rejected here rather than
-     * passed through to EMBL.
+     * Maps chromosome_location to the EMBL /organelle qualifier on the source feature. "Nuclear" is
+     * an allowed value denoting the nuclear/cytoplasmic default and maps to no /organelle qualifier
+     * (handled case-insensitively). All other values are matched case-insensitively against the
+     * INSDC {@code /organelle} controlled vocabulary ({@link ControlledVocabularyUtils.ChromosomeLocation});
+     * this tool now owns validation of that vocabulary, so unrecognised values are rejected here
+     * rather than passed through to EMBL.
      *
      * @param chromosomeLocation  the chromosome location string from the FASTA header
      * @param sourceFt            the source feature to add the qualifier to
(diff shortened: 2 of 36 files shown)


src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/ChromosomeNameFix.java
```diff
@@ -0,0 +1,113 @@
...
+        if (!context.contains(FastaHeaderProvider.class)) {
+            return;
+        }
+
+        String accession = annotation.getAccession();
+        Optional<FastaHeader> headerOpt = context.get(FastaHeaderProvider.class).getHeader(accession);
+        if (headerOpt.isEmpty()) {
+            log.warn("No FASTA header found for accession {}", accession);
+            return;
+        }
```
Rajkumar-D, line 54: FASTA with GFF3 submission should contain headerOpt, should be a hard error when headerOpt is empty

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/FastaHeaderNormalisationFix.java
```diff
@@ -0,0 +1,92 @@
...
+        // No FASTA header source registered for this run -> nothing to fix.
+        if (!context.contains(FastaHeaderProvider.class)) {
+            return;
+        }
+
+        String accession = annotation.getAccession();
+        Optional<FastaHeader> headerOpt = context.get(FastaHeaderProvider.class).getHeader(accession);
+        if (headerOpt.isEmpty()) {
+            log.warn("No FASTA header found for accession {}", accession);
+            return;
```
Rajkumar-D, line 58: Same here, FASTA with GFF3 submission should contain headerOpt, should be a hard error when headerOpt is empty

Rajkumar-D (commented):
Added comment

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/ChromosomeNameFix.java
IvaTutis: resolved with FastaHeaderMappingValidation

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/FastaHeaderNormalisationFix.java
IvaTutis: resolved with FastaHeaderMappingValidation

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
```diff
@@ -73,8 +73,19 @@ protected ValidationEngine initValidationEngine(
         ValidationEngineBuilder builder =
                 new ValidationEngineBuilder().overrideMethodRules(ruleOverrides).failFast(failFast);
 
+        boolean suppliedFastaHeaderProvider = false;
         for (ContextProvider<?> provider : additionalProviders) {
             builder.withProvider(provider);
+            if (provider.type() == FastaHeaderProvider.class) {
+                suppliedFastaHeaderProvider = true;
+            }
```
Rajkumar-D: This solution will work. A cleaner solution is to add `isActive()` to providers (keep it true by default) and use it for checking hasSource in FastaHeaderProvider

-- ContextProvider
```
   /**
     * Returns true if this provider has enough state to be useful on the context.
     * A provider returning false is silently skipped — neither autodetected nor
     * explicitly registered instances of this type land on the context.
     *
     * Override when a provider may be constructed but conditionally empty
     * (e.g. no sources configured). Defaults to true.
     */
    default boolean isActive() {
        return true;
    }
```

FastaHeaderProvider
```
 @Override
    public boolean isActive() {
        return hasSources();
    }
```

--ValidationRegistry
```
 List<ContextProvider<?>> explicitProviders = providers.stream()
                .filter(ContextProvider::isActive)
                .collect(Collectors.toList());

        List<ContextProvider<?>> allProviders = new ArrayList<>();
        if (contextProviderClassPathScanningEnabled) {
            for (ContextProvider<?> provider : instantiateProviders()) {
                if (!provider.isActive()) {
                    continue;
                }
                allProviders.add(provider);
                ctx.register((Class<Object>) provider.type(), (ContextProvider<Object>) provider);
            }
        }
```

Rajkumar-D (commented):
2 comments.
Test was failing and code refactor

src/main/java/uk/ac/ebi/embl/gff3tools/cli/AbstractCommand.java
IvaTutis: I really like this solution! I thought about doing something with whether that there are sources or not, but i didn't consider adding it to the Context Provider interface itself. Elegant!

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/FastaHeaderFormatValidation.java
```diff
@@ -216,12 +214,10 @@ private static void validateChromosomeCombination(FastaHeader header, List<Strin
...
 
-        boolean validCombination = (!hasType && !hasLocation) || (hasName && hasType);
-
-        if (!validCombination) {
-            errors.add("invalid combination of optional chromosome fields - chromosome_type requires "
-                    + "chromosome_name, and chromosome_location requires chromosome_name and chromosome_type "
-                    + "(allowed combinations: none, chromosome_name only, chromosome_name + chromosome_type, "
+        if ((hasType || hasLocation) && !(hasName && hasType && hasLocation)) {
+            errors.add("invalid combination of optional chromosome fields - chromosome_type and "
+                    + "chromosome_location may only be present when chromosome_name, chromosome_type and "
```
Rajkumar-D, line 219: Check if one of chromosome_name, type, location is present then all three should be present.

src/main/java/uk/ac/ebi/embl/gff3tools/gff3toff/GFF3Mapper.java
```diff
@@ -1060,21 +1061,31 @@ private void mapChromosomeType(String chromosomeType, String chromosomeName, Sou
...
+     * case-insensitively against the INSDC {@code /organelle} controlled vocabulary
+     * ({@link ControlledVocabularyUtils.ChromosomeLocation}); this tool now owns validation of that
+     * vocabulary, so unrecognised values are rejected here rather than passed through to EMBL.
      *
      * @param chromosomeLocation  the chromosome location string from the FASTA header
      * @param sourceFt            the source feature to add the qualifier to
      */
     private void mapChromosomeLocation(String chromosomeLocation, SourceFeature sourceFt) {
-        String normalised = chromosomeLocation.toLowerCase();
+        if ("nucleus".equalsIgnoreCase(chromosomeLocation) || "cytoplasm".equalsIgnoreCase(chromosomeLocation)) {
```
Rajkumar-D: We could use the ENUM instead of string

Rajkumar-D (approved):

PR #144, opened 2026-07-01
Move provider initilization to Gff3ProviderFactory so that pipelines could access it

Rajkumar-D:


src/main/java/uk/ac/ebi/embl/gff3tools/Gff3ProviderFactory.java
```diff
@@ -0,0 +1,192 @@
...
+     * </ul>
+     */
+    public static MasterMetadataSource parseMasterEntrySource(Path path) throws ExitException {
+        String ext = getFileExtension(path);
+        return switch (ext) {
+            case "json" -> parseMasterEntryJson(path);
+            case "embl", "ff", "dat" -> parseMasterEntryEmbl(path);
+            default ->
+                throw new CLIException("Unrecognized --master-entry file extension '." + ext
+                        + "'. Supported: .json (MasterEntry JSON), .embl/.ff (EMBL flatfile).");
```
Istar-Eldritch, line 80: We are adding a new extension`.dat`, is this indented? If so, may be worth updating the CLIException to name it.

src/main/java/uk/ac/ebi/embl/gff3tools/Gff3ProviderFactory.java
```diff
@@ -0,0 +1,192 @@
...
+        CompositeSequenceProvider compositeProvider = new CompositeSequenceProvider();
+        for (FileSequenceSource source : sources) {
+            compositeProvider.addSource(source);
+        }
+        return compositeProvider;
+    }
+
+    // ── Private helpers ───────────────────────────────────────────────────────
+
+    private static String getFileExtension(Path path) {
```
Istar-Eldritch, line 126: `AbstractCommand.java` keeps its own `getFileExtension` (returns `Optional<String>`, still used by `resolveSequenceFormat`), and this is a separate private copy (returns `String`, lowercased, empty-string sentinel instead of `Optional.empty()`). The logic is otherwise identical (same `.gz` stripping, same dot-extraction).

Since `Gff3ProviderFactory` already sits at `uk.ac.ebi.embl.gff3tools`, consider elevating this into a small `FileExtensions.getExtension(Path)` helper in the existing `utils/` package (returning `Optional<String>`, lowercased, `.gz`-aware), and have both `AbstractCommand#resolveSequenceFormat` and this method call it. That also lets `AbstractCommand` drop its redundant `.toLowerCase()` call at the resolveSequenceFormat call site.

Istar-Eldritch (approved):

PR #147, opened 2026-07-08
Add GFF3 gap regeneration and rename process command to fix

Istar-Eldritch:
## Summary

Adds GFF3 gap regeneration: discards existing gap/assembly_gap features (SO:0000730) and rebuilds them purely from the runs of N bases in the accompanying FASTA sequence, leaving every other feature untouched.

Concretely this delivers:

1. A new annotation level fix `REGENERATE_GAPS` (`validation/fix/GapRegenerationFix.java`), enabled by default and runtime gated on sequence availability. It runs automatically whenever a `SequenceLookup` is present and is a no op otherwise.
2. A renamed CLI command `process` to `fix` (`cli/FileProcessCommand.java` to `cli/FileFixCommand.java`), which runs the full fix pipeline (including `REGENERATE_GAPS`) over a GFF3 and FASTA pair and writes the fixed GFF3.
3. A refactor of the existing FASTA to GFF3 gap synthesis path used by `conversion` (`fftogff3/FastaToGff3Converter.java`) to delegate to the same `GapRegenerationFix` instead of maintaining a second copy of the gap scanning logic.
4. A shared, extended CLI level gap option validator (`utils/GapOptionsValidator.java`) used by both `fix` and `conversion` for fail fast checking of `--gap-type` and `--linkage-evidence`.
5. A stable (start, end) feature sort on `GFF3Annotation`.

## Motivation

Submitters send GFF3 and FASTA pairs where the GFF3's gap features drift out of sync with the actual N runs in the FASTA (stale coordinates, missing gaps, wrong `estimated_length`). The authoritative signal for a gap is the sequence itself, so the correct fix is to discard the GFF3's gap features and rebuild them from the FASTA. This logic was already used to synthesize gaps during `conversion` (FASTA to GFF3), so this change backs both flows with a single implementation rather than duplicating it.

## Behavior changes worth calling out

1. CLI rename: `process` becomes `fix`, and `-accessions` is dropped in favor of the existing `SequenceOptions` mixin. The old `process` command was a non functional stub (it validated inputs then hit a TODO and wrote nothing), so no automation could have depended on its output.
2. `REGENERATE_GAPS` is enabled by default and sequence gated. This means `conversion` (GFF3 to EMBL) run with `--sequence` will now rebuild gap features from FASTA N runs with default parameters (`min-gap-length=10`, no `gap_type`) instead of passing the GFF3's existing gaps through unchanged. For `validation`, the regenerated features are never written since validation produces no output, so this is harmless there. This default on behavior for `conversion --sequence` is the most consequential change in this PR and is worth an explicit look.

## Design notes

- Gap identification is ontology based (`OntologyClient.findTermByNameOrSynonym` against `SO:0000730`), matching the existing pattern in `GapEstimatedLengthFix` and `AssemblyGapValidation`, so both `gap` and `assembly_gap` (and any synonym) are treated correctly.
- `GapRegenerationFix` is a no-op when no `SequenceLookup` is present, and throws before mutating the annotation if the accession is not covered by any sequence source.
- `AssemblyGapValidation.validGapTypes()` is the single source of truth for the valid `gap_type` vocabulary, used by both the validation rule and `GapOptionsValidator`.
- The `FastaToGff3Converter` refactor is intended to be behavior preserving; existing `fftogff3_rules` fixtures are unchanged and pass.

Full design rationale, alternatives considered, and phased delivery plan are documented in `docs/spec-gap-regeneration.md`.

## Testing

- `GFF3AnnotationTest`: stable ascending (start, end) sort.
- `AssemblyGapValidationTest`: `validGapTypes()` accessor.
- `GapOptionsValidatorTest`: full validation matrix for `--gap-type` and `--linkage-evidence`.
- `GapRegenerationFixTest`: no-op without a sequence, removal and regeneration of gaps, `minGapLength` filtering, `estimated_length`/`gap_type`/`linkage_evidence` attributes, document wide unique IDs across annotations, unknown accession throws before mutation, and feature sorting.
- `FastaToGff3ConverterTest`: updated to register the fix on the engine; existing gap fixtures and expected output are unchanged.
- `FileFixCommandIntegrationTest`: end to end runs of the `fix` command over GFF3 and FASTA pairs, covering gap regeneration, missing sequence, invalid gap options, accession not covered by the FASTA, atomic write behavior, and a case pinning that an invalid FASTA header (e.g. a `molecule_type` outside the controlled vocabulary) fails `fix` with `VALIDATION_ERROR` even though the goal was only gap regeneration. This is expected: `fix` always registers an active `FastaHeaderProvider` because it requires `--sequence`, so the header-format validation added on `main` runs on every `fix` invocation alongside gap regeneration, same as it already does for `conversion`/`validation`.
- `docs/cli-usage-guide.md` updated for the `fix` command and gap options.

Verified locally: `./gradlew spotlessCheck test -Pgitlab_private_token=dummy-token` passes with 1146 tests, 0 failures, 0 errors (after merging `main` and adding the header-validation pinning test above).

### Manual end to end CLI verification

Built the shadow jar and ran `fix` directly against hand built fixtures (two synthetic sequences, one with a stale gap of the wrong length and one missing a gap entirely, plus a non gap feature to confirm it is left untouched):

1. `fix` on the stale GFF3 and FASTA pair: the wrong length gap was corrected, the missing gap was added, and the unrelated feature was preserved. Exit code 0.
2. Same input with `--gap-type` and `--linkage-evidence` supplied: both regenerated gaps carried the expected attributes. Exit code 0.
3. `fix` on the same GFF3 with a FASTA whose header has an invalid `molecule_type`: failed with `FASTA_HEADER_SYNTAX_VALIDATION`, exit code 20, no output file written, matching the pinning test above.

### Known limitation (not addressed in this PR)

`fix` cannot currently repair an input gap feature whose span is not backed by any N run at all (a fully fabricated gap placed over real sequence). The pre-existing `GAP_BASES` check is `FEATURE` level and runs immediately as each feature line is parsed, while `REGENERATE_GAPS` is `ANNOTATION` level and only runs once all of an accession's features have been read, so `GAP_BASES` rejects the bogus feature before `REGENERATE_GAPS` ever gets a chance to remove it. This is a property of when `ValidationType.FEATURE` vs `ValidationType.ANNOTATION` checks run in `GFF3FileReader`, not something `ValidationPriority` can reorder, since priority only orders methods within a single `validate()` call for one target and does not cross that boundary. `fix` still correctly repairs gaps that are merely mis-sized or missing, which covers the motivating case (drifted coordinates); a fully fabricated gap is out of scope here and would need a deliberate follow up (e.g. relaxing `GAP_BASES` specifically inside `fix`'s own engine construction, since `fix` discards and rebuilds every gap regardless of the input's validity).

## Files of interest

- `src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/GapRegenerationFix.java`
- `src/main/java/uk/ac/ebi/embl/gff3tools/utils/GapOptionsValidator.java`
- `src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileFixCommand.java`
- `src/main/java/uk/ac/ebi/embl/gff3tools/fftogff3/FastaToGff3Converter.java`
- `docs/spec-gap-regeneration.md`

[0;32mCommand completed successfully[0m


src/main/java/uk/ac/ebi/embl/gff3tools/cli/FileConversionCommand.java
```diff
@@ -146,6 +143,14 @@ public void run() {
                 sources.add(inputFastaSource);
             }
 
+            // Only the FASTA -> GFF3 branch exposes CLI gap options; every other conversion
+            // direction relies on the default classpath-discovered GapRegenerationFix instance,
+            // which is enabled by default and sequence-gated (no-op without a SequenceLookup).
+            List<Fix> extraFixes = List.of();
+            if (fromFileType == ConversionFileFormat.fasta && toFileType == ConversionFileFormat.gff3) {
+                extraFixes = List.of(new GapRegenerationFix(minGapLength, gapType, linkageEvidence));
+            }
```
Rajkumar-D, line 152: I think this should be handled by FixOverrides, similar to rule overrides, this specific fix will be added from pipeline

src/main/java/uk/ac/ebi/embl/gff3tools/utils/GapOptionsValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Set;
+import uk.ac.ebi.embl.gff3tools.exception.CLIException;
+import uk.ac.ebi.embl.gff3tools.validation.builtin.AssemblyGapValidation;
+
+/**
+ * Fail-fast, CLI-level validation of gap-generation options (--gap-type / --linkage-evidence),
+ * shared by the {@code conversion} (FASTA->GFF3) and {@code fix} commands. Value validity is
+ * defined by {@link AssemblyGapValidation#validGapTypes()} (single source of truth).
+ */
+public final class GapOptionsValidator {
```
Rajkumar-D, line 22: I think we must use ValidationEngine route to validate

PR #155, opened 2026-08-04
Ttena 279 - transl_except attribute value correction

IvaTutis:
Adds the equivalent of Transl_exceptLocationFix.java from sequencetools into gff3tools. 

The additional strand-complement validation makes sense (no complement on non-minus strand in column 7) - verified with bioinformaticians, see gff3 clarifications table.

Relevant as this fix and checks need to execute prior to translation, and transl_except is used there.


Rajkumar-D (approved):
Approved with a comment

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/AttributesLocationValidation.java
```diff
@@ -53,6 +60,48 @@ public void validateTranslExcept(GFF3Annotation gff3Annotation, int line) throws
         validateCodonAttribute(grouped, TRANSL_EXCEPT, line);
     }
 
+    @ValidationMethod(
+            rule = "TRANSL_EXCEPT_STRAND_CONFLICT",
+            description =
+                    "Check that a complement(...) wrapper in a transl_except attribute location agrees with the strand",
+            type = ValidationType.FEATURE,
+            severity = RuleSeverity.ERROR)
```
Rajkumar-D: By default RuleSeverity is ERROR

PR #156, opened 2026-08-07
Ttena 279 - Anticodon attribute fix. 

IvaTutis:
In the anticodon gff3Feature attribute which is inherited from flatfiles, two fixes are done:
- capitalisation of aminoacid in the value of anticodon attribute is corrected
- the subsequence of the sequence is taken out of the FASTA file, if available 

**Argument for keeping this** 

 We currently pass the `anticodon` attribute straight through in both ff<->gff3 conversion directions — nothing in gff3tools reads it, completes it, or normalises it. So a mis-cased aa:SeC stays mis-cased, and a value that arrives without a seq: part leaves without one.

The INSDC standard we write flat files against is specific about that value: aa: has one correct spelling per amino acid, and seq: is the three bases at pos: read on the feature's own strand in lower case — derivable from the sequence, and therefore routinely left out by submitters. We should be filling those in ourselves, because we own the conversion and are the last point at which this can be corrected before we hand out a file we present as INSDC-conformant.

**Change from original behaviour**

Recorded [here](https://embl.atlassian.net/wiki/spaces/TTT/pages/edit-v2/12812319)


IvaTutis:
Closed as the implementation spec changed to match sqtools behaviour

src/main/java/uk/ac/ebi/embl/gff3tools/validation/fix/AnticodonAttributeFix.java
```diff
@@ -0,0 +1,421 @@
...
+
+        return position.hasComplementWrapper();
+    }
+
+    /** @return the reverse complement, or {@code null} if any base is not one the table knows. */
+    private String reverseComplement(String bases) {
+
+        // Anything outside ASCII becomes '?' here, which the table maps to zero like any other
+        // unknown base, so the loop below catches it.
+        byte[] complemented = Translator.reverseComplement(bases.getBytes(StandardCharsets.US_ASCII));
```
Istar-Eldritch: ## Bug: lowercase sequence silently breaks reverse complement on minus strand

`Translator.reverseComplement(byte[])` only maps uppercase letters — its `COMPLEMENT` table (`Translator.java:717`) has entries at indices `'A'`(65) through `'N'`(78), but indices `'a'`(97) through `'n'`(110) are all `0`. The `Translator` itself works around this by calling `toUpperCase()` before `reverseComplement()` internally (`Translator.java:277-280`), but this code bypasses that path.

If `getSequenceSlice` returns lowercase bases, `reverseComplement` produces all-zero bytes, `reverseComplement()` returns `null`, and the `seq:` is **silently not added** for minus-strand features. Forward-strand features are unaffected.

`GFF3Mapper.applySequenceData` (`GFF3Mapper.java:336`) explicitly lowercases the result of `getSequenceSlice`, confirming the return case is not guaranteed to be uppercase.

**Fix:** uppercase `bases` before calling `Translator.reverseComplement`:

```java
byte[] complemented = Translator.reverseComplement(
    bases.toUpperCase(Locale.ROOT).getBytes(StandardCharsets.US_ASCII));
```

The test suite doesn't catch this because all stubs return uppercase (`"TTA"`, `"CAA"`). A test with lowercase input would expose it.

src/main/java/uk/ac/ebi/embl/gff3tools/sequence/SequenceLookup.java
```diff
@@ -63,6 +63,13 @@ public interface SequenceLookup {
     List<GapRegion> getGapRegions(String seqId, long fromBase, long toBase, SequenceRangeOption option)
             throws Exception;
 
+    /**
+     * Whether any of the underlying sources can supply this sequence.=
```
Istar-Eldritch: Minor: stray `=` at end of the Javadoc description — `"...supply this.sequence.="` should be `"...supply this sequence."`

src/test/java/uk/ac/ebi/embl/gff3tools/validation/fix/AnticodonAttributeFixTest.java
```diff
@@ -0,0 +1,437 @@
...
+        GFF3Feature feature = tRna("t1", 1, 100, PLUS, "(pos:complement(500..502),aa:Gln)");
+
+        fix.addSequence(annotation, 1);
+
+        assertEquals(List.of("(pos:complement(500..502),aa:Gln,seq:ttg)"), anticodon(feature));
+    }
+
+    @Test
+    void reverseComplementsOnMinusStrandWithoutComplement() throws Exception {
+        stubSlice("CAA");
```
Istar-Eldritch: Suggestion: add a test that stubs lowercase sequence (e.g. `stubSlice("caa")`) for a minus-strand feature. This would both expose the `reverseComplement` bug above and guard the fix. All existing minus-strand tests use uppercase stubs (`"CAA"`), which is why the bug goes undetected.

Istar-Eldritch (changes requested):
One minor change.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/fix/AnticodonAttributeFixTest.java
IvaTutis: added test

PR #158, opened 2026-08-14
Ttena 209 feature location validation circular fix

IvaTutis:
Fixed what was previously FeatureValidationLocation, and noticed it has an overlap with LocationValidation.  Moved validations from FeatureLengthValidation class into LocationValidation class, so that the location validations relating to feature.end and feature.start are all in the same place. 
Nested 

Logic corrections (arguments [here](https://docs.google.com/document/d/155j8GyWP0ve4m1wRhdEMWZIvJ1RgYVHGgjOCUGlrqMQ/edit?tab=t.0)):
- Corrected for the fact that feature.end on circular sequences is allowed to be greater than sequence length. 
- Corrected the check that feature start<end on feature: _even in circular sequences, as circular sequences (either circular via topology or attribute) are no exception according to the specification (see peter's doc), and they should be of the format  seqLength+endCoordinate, so start<end always hold._


src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidation.java
```diff
@@ -10,52 +10,111 @@
...
 import uk.ac.ebi.embl.gff3tools.gff3.GFF3Annotation;
 import uk.ac.ebi.embl.gff3tools.gff3.GFF3Attributes;
 import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.FastaHeaderProvider;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.utils.ControlledVocabularyUtils;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.utils.FastaHeader;
 import uk.ac.ebi.embl.gff3tools.utils.OntologyClient;
 import uk.ac.ebi.embl.gff3tools.utils.OntologyTerm;
+import uk.ac.ebi.embl.gff3tools.utils.ValidationUtils;
 import uk.ac.ebi.embl.gff3tools.validation.*;
```
Istar-Eldritch: Convention: wildcard imports (`java.util.*`, `validation.*`, `validation.meta.*`) don't match the explicit import style used elsewhere in this file and the AGENTS.md `java, javax, org, com, blank, uk` ordering convention. Spotless doesn't flag them, but worth expanding to explicit imports for consistency.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidation.java
```diff
@@ -10,52 +10,111 @@
...
-    @ValidationMethod(rule = "LOCATION", type = ValidationType.FEATURE)
-    public void validateLocation(GFF3Feature feature, int line) throws ValidationException {
+    @ValidationMethod(rule = RULE_FEATURE_LOCATION_RANGE, type = ValidationType.FEATURE)
+    public void validateLocationRange(GFF3Feature feature, int line) throws ValidationException {
         long start = feature.getStart();
         long end = feature.getEnd();
 
-        if (start <= 0 || end <= 0) {
-            throw new ValidationException(line, INVALID_START_END_MESSAGE.formatted(feature.accession()));
+        if (end < start) {
```
Istar-Eldritch: Removing the circular exemption on `end < start` is correct given the redesign: circular wrap-around is now represented as `end = physicalEnd + seqLen` (still `end > start`), validated by the new `isCircular` branch in `validateFeatureEndWithinSequence`. Worth confirming with whoever owns the FF-to-GFF3 writer that no producer still emits circular features as `start > end` without the offset, since this rule now rejects that shape.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidation.java
```diff
@@ -10,52 +10,111 @@
...
         }
-        boolean isCircular = Boolean.TRUE
-                .toString()
-                .equalsIgnoreCase(
-                        feature.getAttribute(GFF3Attributes.CIRCULAR_RNA).orElse("false"));
-        if (!isCircular && end < start) {
-            throw new ValidationException(line, INVALID_START_END_MESSAGE.formatted(feature.accession()));
+    }
+
+    @ValidationMethod(
```
Istar-Eldritch: Soundness: priority regression. The old `FeatureLocationValidation` explicitly set `priority = ValidationPriority.LOW` on both rules. These merged methods carry no `priority` argument, so they silently run at `NORMAL` now. `ValidationEngine` executes validations in priority tiers and gates fail-fast per tier (`ValidationEngine.java:61-63`), so this changes ordering and fail-fast behaviour. If the demotion wasn't deliberate, add `priority = ValidationPriority.LOW` back on `validateFeatureEndWithinSequence`, and consider whether the newly split below-one and range rules should match, since they descend from the same LOW-priority contract.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidation.java
```diff
@@ -116,4 +175,41 @@ public void validateCdsLocation(GFF3Annotation annotation, int line) throws Vali
...
+     * a missing mandatory topology is reported by {@link FastaHeaderFormatValidation}.
+     */
+    private boolean isCircularSequence(String accession) {
+        if (!context.contains(FastaHeaderProvider.class)) {
+            return false;
+        }
+        return context.get(FastaHeaderProvider.class)
+                .getHeader(accession)
+                .map(FastaHeader::getTopology)
+                // Canonicalise rather than matching the raw value: FastaHeaderNormalisationFix is
```
Istar-Eldritch: Soundness: the comment asserts `FastaHeaderNormalisationFix` runs only after this annotation's features are validated. I didn't verify that ordering against the fix's actual `ValidationType`/priority. It is load-bearing for why raw (uncanonicalised) topology values can reach this code, so worth a second look.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidationTest.java
```diff
@@ -11,21 +11,32 @@
...
 import uk.ac.ebi.embl.gff3tools.gff3.GFF3Attributes;
 import uk.ac.ebi.embl.gff3tools.gff3.GFF3Feature;
+import uk.ac.ebi.embl.gff3tools.sequence.SequenceLookup;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.FastaHeaderProvider;
+import uk.ac.ebi.embl.gff3tools.sequence.fasta.header.utils.FastaHeader;
 import uk.ac.ebi.embl.gff3tools.utils.OntologyTerm;
+import uk.ac.ebi.embl.gff3tools.validation.ContextProvider;
+import uk.ac.ebi.embl.gff3tools.validation.ValidationContext;
 
 public class LocationValidationTest {
```
Istar-Eldritch: Convention: AGENTS.md calls for package-private test classes (no `public` modifier). JUnit 5 doesn't require `public`. Worth dropping the modifier for convention adherence.

src/main/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidation.java
Istar-Eldritch: The old `FeatureLocationValidation` explicitly set `priority = ValidationPriority.LOW` on both rules. These merged methods carry no `priority` argument, so they silently run at `NORMAL` now, so this changes ordering and fail-fast behaviour. If the demotion wasn't deliberate, add `priority = ValidationPriority.LOW` back on `validateFeatureEndWithinSequence`, and consider whether the newly split below-one and range rules should match, since they descend from the same LOW-priority contract.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidationTest.java
Istar-Eldritch: A nit, the convention is for  package-private test classes (no `public` modifier). JUnit 5 doesn't require `public`. Worth dropping the modifier for convention adherence.

Istar-Eldritch (changes requested):
### GFF3-to-FF (GFF3 to EMBL): validation passes, mapper silently writes wrong coordinates

The validation fix is the point of the PR for this direction: GFF3 files using `end = physicalEnd + seqLen` for circular origin-spanning features now pass validation instead of failing. That unblocks conversion of these files.

But `GFF3Mapper.mapGFF3Location` (`GFF3Mapper.java:271-288`) passes the raw GFF3 `start` and `end` straight into `locationFactory.createLocalRange(start, end)` with no logic to unwrap `end > seqLen` into the EMBL compound-location form (`join(149000..149696,1..966)`). So conversion now succeeds but writes `149000..150662` to the EMBL output, which is an invalid location for a 149696bp sequence.

Before this PR, validation would reject the feature and block conversion. After this PR, validation passes and the mapper silently produces an EMBL file with incorrect coordinates.

The mapper needs to handle `end > seqLen` by splitting into a compound location.  No test resource has a feature where end > seqLen on a circular molecule. The GFF3ToFFConverterTest auto-discovers all .gff3/.embl pairs in gff3toff_rules/, so adding a new pair like circular_origin_spanning.gff3 + circular_origin_spanning.embl would be the natural way to cover this.

src/test/java/uk/ac/ebi/embl/gff3tools/validation/builtin/LocationValidationTest.java
IvaTutis: done, also did the cleanup of walking through the rest of public test classes and removed the public modifier. Out of the scope of the PR, but should have been done anyway.

PR #159, opened 2026-08-14
Ttena 279 edge n validation

IvaTutis:
Validation for linear sequences that confirms they have no edge Ns. 
I also took the liberty to simplify the SequenceLookup interface, as there is no need to have SequenceRangeOption's anymore as we are not trimming Ns


Istar-Eldritch (approved):
🚀 brill PR, I was considering adding warnings when a provider is not initialized and a validation that depends on it runs instead of skipping silently, but I think that falls out of the scope of this PR.

